PR #2, opened 2022-07-14
Feat/cli add tests apps 617

pawartur:
# Tests for the `composedb` `cli` package

## Description

Tests were moved and adapted from `js-glaze`'s `feat/composites` branch. The test for `composite:deploy` was commented out, because we can't make it work reliably for now. The idea to solve this is to enable a config option for the ceramic daemon that will make it run in an isolated mode, not connecting to other p2p ipfs nodes.

The tests checking that `graphql:*` commands succeed where initially added here, but then removed, because the commands started failining with unexpected timeouts. There will be another PR shortly fixing the commands and re-adding the tests.

## How Has This Been Tested?

- [X] Ran all the tests

## Definition of Done


packages/cli/globalTeardown.js
```diff
@@ -0,0 +1,5 @@
+import { teardown } from 'jest-dev-server'
+
+export default async function globalTeardown() {
+  await teardown()
```
PaulLeCam, line 4: Maybe part of the teardown could be to delete the `CONFIG_DIR_PATH` folder created in the setup script?

PaulLeCam (dismissed):
LGTM just wondering if we can use the teardown script to do additional cleanup, but no need to focus on this now, would just be a nice to have.

packages/cli/globalTeardown.js
pawartur: @PaulLeCam , I added the removal of the whole `TEMP_DIR_PATH` whose `CONFIG_DIR_PATH` is a subdirectory: https://github.com/ceramicstudio/js-composedb/pull/2/files#diff-2eee76cf110a51a727cb612325f04406c422d0b3891d1f9bd8478abfafbc94cbR6

Does it work for you :) ?

.npmrc
```diff
@@ -0,0 +1 @@
+auto-install-peers=true
```
PaulLeCam, line 1: Can you please remove this?

packages/cli/globalConsts.js
```diff
@@ -0,0 +1,10 @@
+import process from 'node:process'
```
PaulLeCam, line 1: Not an issue but this shouldn't be needed, node injects `process` automatically

packages/cli/globalSetup.js
```diff
@@ -0,0 +1,57 @@
+import { setup } from 'jest-dev-server'
+import { Model } from '@ceramicnetwork/stream-model'
+import fs from 'fs-extra'
+import './globalConsts.js'
```
PaulLeCam, line 4: Why is this imported like this?

packages/cli/globalSetup.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default async function globalSetup() {
+  if (await fs.exists(TEMP_DIR_PATH)) {
+    await fs.rm(TEMP_DIR_PATH, { force: true, recursive: true })
+  }
+
+  await fs.ensureDir(TEMP_DIR_PATH)
+  await fs.ensureDir(TEST_OUTPUT_DIR_PATH)
+  await fs.ensureDir(CONFIG_DIR_PATH)
+  await fs.ensureDir(STATE_STORE_DIR_PATH)
```
PaulLeCam, line 47: Could these be run in parallel?

packages/cli/src/commands/document/content.ts
```diff
@@ -39,6 +41,7 @@ export default class DocumentContent extends Command<DocumentContentFlags, { str
         this.log(JSON.stringify(mid.content))
       }
     } catch (e) {
+      console.error(e)
```
PaulLeCam, line 44: Extra `console.error`

packages/cli/test/composites.test.ts
```diff
@@ -0,0 +1,317 @@
...
+        `${dirpath}/${filename}.js`,
+        `${dirpath}/${filename}.ts`,
+      ])
+      expect(
+        compileWithJustCompositePath.stderr.toString().includes('Compiling the composite... Done!')
+      ).toBe(true)
+
+      const jsonRepresentation = await fs.readFile(`${dirpath}/${filename}.json`)
+      const jsRepresentation = await fs.readFile(`${dirpath}/${filename}.js`)
+      const tsRepresentation = await fs.readFile(`${dirpath}/${filename}.ts`)
```
PaulLeCam, line 310: Could these be loaded in parallel?

PaulLeCam (changes requested):

packages/cli/globalSetup.js
```diff
@@ -0,0 +1,57 @@
...
+  await fs.writeJson(CONFIG_PATH, TEST_DAEMON_CONFIG)
+
+  await setup({
+    command:
+      `CERAMIC_ENABLE_EXPERIMENTAL_INDEXING=\'true\' rm -rf ~/.goipfs && pnpm dlx @ceramicnetwork/cli@^2.4.0 daemon --config ${CONFIG_PATH}`,
```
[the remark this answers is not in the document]
pawartur: I removed the hard-coded version and will see, if there's a way to tell the `daemon` where it should create the `.goipfs` dir. Thanks, @PaulLeCam :) !

PaulLeCam (approved):
Great, thanks for the changes!

PR #6, opened 2022-07-25
Fix typo

oed:


website/docs/first-composite.mdx
```diff
@@ -121,7 +121,7 @@ composedb composite:models my-first-composite.json
 - Edit config file to add model ID to index
 - Restart Ceramic node
 
-## Interacting using GraphiQL
+## Interacting using GraphQL
```
youngnishant, line 124: Graphiql is the tool used, so its correct, there is no typo mistake. Find picture attached as a proof
Screenshot 2022-07-25 at 9 38 09 PM

oed: oh, my bad!

PR #13, opened 2022-08-10
Bug/graphql server timeouts apps 640

pawartur:
# Fixed the `graphql:server` command timing out after a change in `oclif/core`

## Description

`oclif` started automatically killing all the commands 10s after their `run()` method exits (https://github.com/oclif/oclif.github.io/pull/166/files, more context on the preferred solution to this in https://github.com/oclif/core/issues/464).

The solution upvoted by one of the devs is to prevent the `run()` method from exiting, so that's what this PR introduces.

## How Has This Been Tested?

- [X] Re-added the tests for `graphql:server` command which are now passing

## Definition of Done

### DIFF
--- a/packages/cli/src/commands/graphql/server.ts
+++ b/packages/cli/src/commands/graphql/server.ts
@@ -56,11 +56,13 @@ export default class GraphQLServer extends Command<
         port: this.flags.port,
       })
       this.spinner.info(`GraphQL server is listening on ${handler.url}`)
-      process.on('SIGTERM', () => {
+      const handleProcessKillGracefully = () => {
         handler.stop(() => {
           this.spinner.succeed('Server stopped')
         })
-      })
+      }
+      process.on('SIGTERM', handleProcessKillGracefully)
+      process.on('SIGINT', handleProcessKillGracefully)
       // oclix/core has started to force kill all the commands after 10s, without an option to change this behaviour
       // Issue with more context: https://github.com/oclif/core/issues/464
       // Update in their docs making it clear (in response to the issue above): https://github.com/oclif/oclif.github.io/pull/166/files
--- a/packages/cli/src/commands/graphql/server.ts
+++ b/packages/cli/src/commands/graphql/server.ts
@@ -61,6 +61,10 @@ export default class GraphQLServer extends Command<
           this.spinner.succeed('Server stopped')
         })
       })
+      // oclix/core has started to force kill all the commands after 10s, without an option to change this behaviour
+      // Issue with more context: https://github.com/oclif/core/issues/464
+      // Update in their docs making it clear (in response to the issue above): https://github.com/oclif/oclif.github.io/pull/166/files
+      await new Promise(() => undefined)
     } catch (e) {
       this.spinner.fail((e as Error).message)
       return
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/graphql/server.ts
```diff
@@ -61,6 +61,10 @@ export default class GraphQLServer extends Command<
           this.spinner.succeed('Server stopped')
         })
       })
+      // oclix/core has started to force kill all the commands after 10s, without an option to change this behaviour
+      // Issue with more context: https://github.com/oclif/core/issues/464
+      // Update in their docs making it clear (in response to the issue above): https://github.com/oclif/oclif.github.io/pull/166/files
+      await new Promise(() => undefined)
```
PaulLeCam, line 67: It seems this would keep the process hanging forever, no? Does `SIGTERM` still terminates the process as expected?

pawartur: Yes, I ran the command:
```
☁  cli [bug/graphql-server-timeouts-APPS-640] node bin/run.js graphql:server /Users/arturwdowiarski/src/js-pinner-analytics/data/runtime.merged.composite.json
ℹ GraphQL server is listening on http://localhost:65426/graphql

```
, then grepped the process id and killed the process:
```
☁  js-composedb [bug/graphql-server-timeouts-APPS-640] ps aux | grep run.js                                                                                                         
arturwdowiarski  55244   0.0  0.0 408637584   1792 s003  S+    5:21pm   0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn run.js
arturwdowiarski  55242   0.0  0.4 420043728 122352 s002  S+    5:21pm   0:00.82 node bin/run.js graphql:server /Users/arturwdowiarski/src/js-pinner-analytics/data/runtime.merged.composite.json
☁  js-composedb [bug/graphql-server-timeouts-APPS-640] kill 55242
```
, then checked that the command exited gracefully:
```
ℹ GraphQL server is listening on http://localhost:65426/graphql
✔ Server stopped
```
, then checked that the process was no longer running:
```
☁  js-composedb [bug/graphql-server-timeouts-APPS-640] ps aux | grep run.js
arturwdowiarski  55311   0.0  0.0 408628368   1648 s003  S+    5:23pm   0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn run.js
```

PaulLeCam: Does it exit as expected on control+C or do you need to explicitly kill it like this?

pawartur: It works with ctrl+C too :)

PaulLeCam: OK perfect, thanks for checking!

PaulLeCam (approved):

packages/cli/src/commands/graphql/server.ts
pawartur: But ctrl+C sends a SIGINT, not SIGTERM and we don't handle this gracefully now. I can add handling for SIGINT too.

PaulLeCam (approved):

PR #15, opened 2022-08-12
Use `ModelInstanceDocument.single()` for deterministic documents

PaulLeCam:
Needed some change to the runtime composite definition to support the logic, other than this it's pretty self-contained.

Opening as draft until the stable Ceramic packages are released and used in this PR.


PaulLeCam:
@pawartur should be ready for review now, if you can have a look please?
Part of the changes I update TypeScript to v4.8 which as stricter type checks, so I had to change some unrelated parts of the code that started to fail TS checks.

packages/client/src/context.ts
```diff
@@ -71,7 +71,7 @@ export class Context {
   /**
    * Load a document by ID, using the cache if possible.
    */
-  async loadDoc<Content = Record<string, any>>(
+  async loadDoc<Content extends Record<string, any> = Record<string, any>>(
```
pawartur, line 74: Are we using the possibility to extend `Record<sting, any>` anywhere?

PaulLeCam: It's needed with TS 4.8 being more strict about types, as `TileDocument` defines its contents as `Record<string, any>`, we need to specify the `Content` type here implements this type.

pawartur: Thanks :)

PR #16, opened 2022-08-15
Prepare v0.2

PaulLeCam:
Updates the website setup to support versions, fill-in some contents and bump packages version.


website/docusaurus.config.js
```diff
@@ -63,11 +70,18 @@ const config = {
...
             activeBasePath: 'docs/api',
             position: 'left',
             label: 'API',
           },
+          // {
+          //   type: 'docsVersionDropdown',
+          //   position: 'right',
+          //   dropdownItemsAfter: [{ to: '/versions', label: 'All versions' }],
+          //   dropdownActiveClassDisabled: true,
+          // },
```
pawartur, line 84: Did you leave it here on purpose, @PaulLeCam ?

pawartur (approved):
Thanks, @PaulLeCam , looks good!

Two thoughts I had while reviewing:
* would it make sense to make the changelog a bit more descriptive? I think that it might be useful even for us in the future
* to me personally the two main categories we have - Docs and API - are confusing, as all of the contents are docs in broad sense and the API sections contains mostly reference docs. I'd rename the sections maybe to Usage and Docs respectively? Just a thought.

website/docusaurus.config.js
PaulLeCam: Yes, that will be needed once we add a new version.

PaulLeCam:
>     * would it make sense to make the changelog a bit more descriptive? I think that it might be useful even for us in the future

Yeah I expect in the future we can have more descriptive info, I started using https://github.com/changesets/changesets so it's mostly getting familiar with the tool.

>     * to me personally the two main categories we have - Docs and API - are confusing, as all of the contents are docs in broad sense and the API sections contains mostly reference docs. I'd rename the sections maybe to Usage and Docs respectively? Just a thought.

I mostly followed the conventions from https://docusaurus.io, https://nextjs.org/ and similar. We could adopt other conventions but maybe we should define these at some point then, especially if they differ from widely-used projects?

PR #20, opened 2022-08-23
Bug/failing composite from apps 667

pawartur:
# Remove CompositeInput type used to wrap `CompositeParams` in `new Composite(compositeParams)`.

## Description

The cause for creating this PR was a bug described in the section below.

This PR removes the `CompositeInput` type which made it possible not to wrap `CompositeParams` with `new Composite(compositeParams)` when calling certain methods on `Composite`. This possibility wasn't actually used in the code, though.

Removing `CompositeInput` made it possible to also remove `instanceof` checks in a couple of internal methods, which should help avoid certain bugs related to `instanceof` being a bit too strict in javascript (like the bug described below).

### How to reproduce

To Reproduce:
1. Remove `pnpm` completely from your system
2. Install `pnpm`
3. Install ComposeDB CLI globally with `pnpm: pnpm add -g @composedb/cli@^0.2.0`
4. Download the first composite schema used in cli tests from https://github.com/ceramicstudio/js-composedb/blob/main/packages/cli/test/scripts/data/composite.picture.post.schema
5. Download the second composite schema used in cli tests from https://github.com/ceramicstudio/js-composedb/blob/main/packages/cli/test/scripts/data/composite.profiles.schema
6. Start running a local ceramic node on my machine: `ERAMIC_ENABLE_EXPERIMENTAL_INDEXING='true' pnpm dlx @ceramicnetwork/cli daemon --network=inmemory`
7. Create a composite out of the first schema: `composedb composite:create composite.profiles.schema -o encoded.composite.profiles.json -k 7abdadac0eaf13799f6efb326fc6d4d0e27e28a38305356856545302b063f08c`
8. Create a composite out of the second schema: composedb composite:create composite.picture.post.schema -o encoded.composite.picture.post.json -k 7abdadac0eaf13799f6efb326fc6d4d0e27e28a38305356856545302b063f08c 
9. Try to merge the two composites: `composedb composite:merge encoded.composite.picture.post.json  encoded.composite.profiles.json`

Result:
```
✖ Cannot read properties of undefined (reading 'version')
```

Expected Result:
You should be able to merge the two composites successfully

### The Cause

I figured out that this error is being thrown after the composite:merge command loads both composites from input paths and passes them in an array to Composite.from(…), then the instanceof check fails [here](https://github.com/ceramicstudio/js-composedb/blob/main/packages/devtools/src/composite.ts#L282), the function tries to create a Composite instance passing a valid Composite instance to the constructor and then the constructor fails, because of a missing param.definition.version key.

Did some debugging to check why the `instanceof` check might be failing and figured out that even with this simple pnpm setup with just one global package being composedb/cli I somehow have two devtools packages installed:
```
☁  pnpm  find . | grep composite.js
./global/5/.pnpm/@composedb+devtools@0.2.0/node_modules/@composedb/devtools/dist/composite.js
./global/5/.pnpm/@opentelemetry+core@1.2.0/node_modules/@opentelemetry/core/build/esm/propagation/composite.js.map
./global/5/.pnpm/@opentelemetry+core@1.2.0/node_modules/@opentelemetry/core/build/esm/propagation/composite.js
./global/5/.pnpm/@opentelemetry+core@1.2.0/node_modules/@opentelemetry/core/build/esnext/propagation/composite.js.map
./global/5/.pnpm/@opentelemetry+core@1.2.0/node_modules/@opentelemetry/core/build/esnext/propagation/composite.js
./global/5/.pnpm/@opentelemetry+core@1.2.0/node_modules/@opentelemetry/core/build/src/propagation/composite.js.map
./global/5/.pnpm/@opentelemetry+core@1.2.0/node_modules/@opentelemetry/core/build/src/propagation/composite.js
./global/5/.pnpm/@composedb+devtools@0.2.0_graphql@16.6.0/node_modules/@composedb/devtools/dist/composite.js
...
```

I checked it further and noticed that `./global/5/.pnpm/@composedb+devtools@0.2.0_graphql@16.6.0/node_modules/@composedb/devtools/dist/composite.js` is actually a symlink to `./global/5/.pnpm/@composedb+devtools@0.2.0/node_modules/@composedb/devtools/dist/composite.js`, but it still makes the `instanceof` check fail, as it doesn't care about symlinks, only about paths.

## How Has This Been Tested?

- [X] Updated and ran the tests

## Definition of Done

- [X] I have updated the READMEs of affected packages
- [X] I have made corresponding changes to the documentation
- [X] The changes have been communicated to interested parties


ukstv (approved):
What is the way now to merge multiple `CompositeInput`s?

PaulLeCam (changes requested):
Please, let's not jump the gun on failing `instanceof` checks, these are setup issues not code issues.
Happy to discuss next week when I'm back but let's not make drastic changes now.

pawartur:
> What is the way now to merge multiple `CompositeInput`s?

In this PR? As there's no `CompositeInput` type anymore, you need to explicitly use `new Composite(compositeParams)` when merging multiple composites, rather than just pass the `compositeParams`.

PaulLeCam:
@pawartur can we close this if fixed with the dependencies setup change please?

pawartur:
Yes, I'll close this, @PaulLeCam , thanks. Will also create a reminder for us to double-check the dependencies in this repo. I'm worried that moving `devtools` from `dependencies` to `peerDependencies` may be hiding a deeper issue.

PR #22, opened 2022-08-25
README.md

yalexis-eth:
Updated the README


3avi:
sorry for the late reply but thank you for the wording suggestion yalexis!

PR #32, opened 2022-09-14
Add null checks to query results handling

PaulLeCam:
There are a bunch of unrelated changes dues to linting fixes and dependencies updates, but the main change is adding null-checks to the client handling of query results to ensure missing nodes don't result in runtime errors.
This does not change the high-level GraphQL API as the generated schema already treats these fields as possibly null.


pawartur (approved):
Thanks!

PR #33, opened 2022-09-15
Updated ceramic deps

pawartur:
# Updated ceramic deps and updated the /client and /cli packages to accommodate null stream states returned by the queryAPI

## Description

In the latest RCs of http-client and common the change was introduced that make it possible for the queryAPI to return null nodes in paginated responses. This PR updates the /cli and /client packages to accomodate that change.

Didn't write any additional tests for that. Can do it in this or in the next PR, depending on the reviews.

## How Has This Been Tested?

- [X] Ran all the existing tests
- [X] Manually tested the case when a model requested from index is missing from the state store:
1. Started the ceramic daemon
2. Used composedb cli to create a model from composite SDL definition
3. Stopped the ceramic daemon
4. Deleted the statestore dir used by the ceramic daemon
5. Started the ceramic daemon again
6. Tried to list the model
```
☁  cli [chore/version-bump] node bin/run.js model:list --indexer-url=http://localhost:7007
{"id":"<MISSING!!>","name":"<MISSING!!>","description":"<MISSING!!>"}

✔ Loading models... Done
```

## Definition of Done

- [] I have added relevant tests for new or updated functionality

### DIFF
--- a/examples/next-notes/package.json
+++ b/examples/next-notes/package.json
@@ -10,7 +10,7 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@ceramicnetwork/http-client": "^2.3.4",
+    "@ceramicnetwork/http-client": "^2.3.5-rc.2",
     "@composedb/client": "workspace:^0.2.0",
     "@composedb/types": "workspace:^0.2.0",
     "@emotion/react": "^11.10.0",
--- a/packages/cli/src/commands/model/list.ts
+++ b/packages/cli/src/commands/model/list.ts
@@ -58,7 +58,7 @@ export default class ModelList extends BaseCommand<ModelListFlags> {
       while (this.lastLoadedPageInfo?.hasNextPage) {
         await CliUx.ux.anykey('Press any key to load more models')
         this.spinner.start('Loading models...')
-        const nextPage: Page<StreamState> = await ceramicIndexer.index.queryIndex({
+        const nextPage: Page<StreamState | null> = await ceramicIndexer.index.queryIndex({
           first: this.getPageSize(),
           model: Model.MODEL,
           after: this.lastLoadedPageInfo?.endCursor,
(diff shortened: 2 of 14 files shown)


examples/next-notes/package.json
```diff
@@ -10,7 +10,7 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@ceramicnetwork/http-client": "^2.3.4",
+    "@ceramicnetwork/http-client": "^2.3.5-rc.2",
```
PaulLeCam, line 13: Don't bump to RCs or other non-stable versions please, that will create duplicate dependencies for consumers.

packages/cli/src/commands/model/list.ts
```diff
@@ -74,14 +74,23 @@ export default class ModelList extends BaseCommand<ModelListFlags> {
...
-        description: (stream.content as Record<string, any>).description as string,
+      if (edge && edge.node) {
+        const stream = ceramicIndexer.buildStreamFromState(edge.node)
+        return {
+          id: stream.id.toString(),
+          name: (stream.content as Record<string, any>).name as string,
+          description: (stream.content as Record<string, any>).description as string,
+        }
+      } else {
+        return {
```
PaulLeCam, line 87: Why empty contents here please? Let's filter out null results rather than returning them please.

PaulLeCam (changes requested):
I already addressed the changes for the client in https://github.com/ceramicstudio/js-composedb/pull/32 can we merge it first please?
Also, please don't use RC versions, we are not publishing RCs for ComposeDB packages and shouldn't mix them with regular releases.

packages/cli/src/commands/model/list.ts
pawartur: Are you ok with pagination working in unexpected ways if we filter out empty results, @PaulLeCam

PaulLeCam: How does it affect pagination? We can always perform multiple queries to fill a page if needed, IMO it's better than empty rows.

pawartur: I guess you answered your own question about the pagination :). Cool, thanks!

pawartur:
Closing this PR, as I wasn't aware of https://github.com/ceramicstudio/js-composedb/pull/32 which did most of the changes from here and was already merged. Also, we need to update the ceramic packages again (to non-rc versions), before bumping deps in composedb packages, so that we don't refer RC versions in deps.

Will create a new PR once ceramic packages are updated again.

PR #35, opened 2022-09-23
Simple relations support

PaulLeCam:
Will need more integration tests once queries are supported on Ceramic's side, but should be OK to start reviewing as-is.


packages/client/src/schema.ts
```diff
@@ -385,25 +409,119 @@ class SchemaBuilder {
...
+            if (id == null) {
+              return null
+            }
+            const loaded = await ctx.loadDoc(id)
+            if (loaded == null) {
+              return null
+            }
+            const loadedModel = loaded.metadata.model.toString()
+            if (relation.model != null && loadedModel !== relation.model) {
+              console.warn(
```
pawartur, line 445: When can this happen?

pawartur (approved):
I feel I can't reliably review such a huge PR, but I have read through the changes and didn't notice anything that would look problematic.

packages/client/src/schema.ts
PaulLeCam: That could happen if the server doesn't perform this check when creating or querying relations, this ensures it doesn't create a situation the client couldn't handle.

PR #51, opened 2022-10-19
Consistent and user-friendly message when stopping model load manually

pawartur:
# Consistent and user-friendly message when stopping model load after a page is loaded

## How Has This Been Tested?

- [X] Manually tested that the UX for stopping the command is the same for ctrl+c and q press:

Screenshot 2022-10-20 at 11 35 22

## Definition of Done

### DIFF
--- a/packages/cli/src/commands/model/list.ts
+++ b/packages/cli/src/commands/model/list.ts
@@ -60,8 +60,8 @@ export default class ModelList extends BaseCommand<ModelListFlags> {
       while (this.lastLoadedPageInfo?.hasNextPage) {
         this.spinner.stop()
         await this.anykeyWithFriendlyExit(
-          'Press any key to load more models',
-          'You stopped the model loading manually'
+          'Press ctrl+c or q to quit. Press any other key to continue loading models.',
+          'Stopped loading models due to user input'
         )
         this.spinner.start('Loading models...')
         const nextPage: Page<StreamState | null> = await ceramicIndexer.index.query({
--- a/packages/cli/src/commands/model/list.ts
+++ b/packages/cli/src/commands/model/list.ts
@@ -59,7 +59,10 @@ export default class ModelList extends BaseCommand<ModelListFlags> {
 
       while (this.lastLoadedPageInfo?.hasNextPage) {
         this.spinner.stop()
-        await CliUx.ux.anykey('Press any key to load more models')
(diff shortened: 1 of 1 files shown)


PaulLeCam (dismissed):

pawartur:
> @pawartur I have a question. When is "You stopped the model loading manually" message displayed? When there are no more pages, or when a user hits Ctrl+C?

It's displayed when a user get the "Press any key to load more" prompts, but they press "q" or "ctrl"+"c".

It's still not great, because, if a user presses "ctrl"+"c" while the models are loaded, the UX is still different, but this is something I'd like to fix in the next step.

ukstv (dismissed):

ukstv:
Lovely! Are you open to an idea to involve adoption to provide some input on wording @pawartur ?

pawartur:
> Lovely! Are you open to an idea to involve adoption to provide some input on wording @pawartur ?

Absolutely! I think it's a great idea! Thanks, @ukstv . Adding @mattdavis0351 and @Sterahi as reviewers :).

pawartur:
@mattdavis0351 , @Sterahi , could you have a look at the DX proposed here and let us know your thoughts?

Sterahi:
> press any key to load mode models:

With that being the flow that we're using I do see a few changes that could make the user experience better.
1. `Press any key to load more models` would make more sense (to me) as `Press any key to continue loading models.`
2. I'd like to see a prompt that users could cancel the loading with either `ctrl+c` or `q`. Something like `ctrl+c or q to stop loading model`.
3. The message when manually quitting is a little unclear, to me it would make more sense as `Stopped loading models due to user input` or something similar but that is just a _very_ minor complaint.

pawartur:
> > press any key to load mode models:
> 
> With that being the flow that we're using I do see a few changes that could make the user experience better.
> 
> 1. `Press any key to load more models` would make more sense (to me) as `Press any key to continue loading models.`
> 2. I'd like to see a prompt that users could cancel the loading with either `ctrl+c` or `q`. Something like `ctrl+c or q to stop loading model`.
> 3. The message when manually quitting is a little unclear, to me it would make more sense as `Stopped loading models due to user input` or something similar but that is just a _very_ minor complaint.

Thank You! I'll apply these suggestion in this PR :)

ukstv (approved):

Issue #53, opened 2022-10-28
Ensure DateTime scalar inputs are converted to string


PaulLeCam:
# Description

The [`DateTime` scalar from the `graphql-scalars` library](https://www.the-guild.dev/graphql/scalars/docs/scalars/date-time) converts inputs to JS Date objects rather than strings as expected.
There is a [PR already open to fix the issue](https://github.com/Urigo/graphql-scalars/pull/1641) but we should address it directly here in the meantime.

# Technical Information

The [PR to fix the issue in `graphql-scalars`](https://github.com/Urigo/graphql-scalars/pull/1641) could a good solution to look into, implementing similar changes in our [`@composedb/graphql-scalars` package](https://github.com/ceramicstudio/js-composedb/tree/main/packages/graphql-scalars).

morozj01:
@PaulLeCam 

It seems like [@composedb/graphql-scalars](https://github.com/ceramicstudio/js-composedb/blob/ca5c9b60e758d7865aec8d72f5fc94ecd7717168/packages/graphql-scalars/src/index.ts) is implemented as a simple pass through of the scalar types defined in the [graphql-scalars](https://www.npmjs.com/package/graphql-scalars) package.

The PR you referenced is fixing an issue in the `serialize` function of the datetime scalar which that package is exporting

I could implement a custom definition for the datetime [scalar type](https://graphql.org/graphql-js/type/#graphqlscalartype) (to use temporarily in place of the one being exported by [graphql-scalars](https://www.npmjs.com/package/graphql-scalars)) but that may be overkill for a package which is currently very simple - especially with a PR pending with a fix.

Happy to work on a temporary solution but wanted to get your feedback before proceeding.

stbrody:
Shouldn't dates get converted into number types, or ideally to native date objects in the underlying db, so that range queries can work effectively?

stbrody:
I guess the problem is that JSON doesn't have a native encoding for dates.  Sigh, this was one of the reasons that MongoDB's BSON added extra types beyond the base supported in JSON.  We might want to consider defining a JSON-compatible serialization format for dates so that Ceramic nodes can actually treat dates like dates

morozj01:
Exactly - ceramic is importing a library which is supposed to serialize dates into strings already but there is a bug where it is not being done.

They have a PR open with a potential fix but it has been inactive for a while

https://github.com/ceramicstudio/js-composedb/pull/54
[...]

PR #54, opened 2022-11-07
fix: implement temporary schema definition to support datetime scalars

morozj01:
# Ensure DateTime scalar inputs are converted to string - #53 

## Description
This is meant to be a temporary solution to enable datetime scalar types in Ceramic models while [this pull request](https://github.com/Urigo/graphql-scalars/pull/1641) is in review (as there has been no activity for some time.)

This change introduces a custom datetime scalar definition in a new file which is meant to be deleted after the solution is merged into the [graphql-scalars package](https://www.npmjs.com/package/graphql-scalars)

This schema definition is essentially a simplified copy of the logic defined in the [graphql-scalars package](https://github.com/Urigo/graphql-scalars/blob/master/src/scalars/iso-date/DateTime.ts).

It purposely kept minimal and therefore only ISO strings in UTC time are supported - as per the [documentation](https://composedb.js.org/docs/0.3.x/guides/creating-composites/scalars#datetime)

Passing in Unix timestamp or JS date object will throw an error.

## How Has This Been Tested?

Currently creating documents with a datetime scalar field will throw [this error](https://forum.ceramic.network/t/cbor-encode-error-unsupported-type-date-when-trying-to-create-composedb-document/360/3)

With this change, I was able to upload datetime scalar values without issue

IDs: 
- kjzl6kcym7w8y5f7jspauwxecxeuzrri0rigwwl5h7x8i9ebhmh7hi2thxmksx0
- kjzl6kcym7w8y60p5d46t282vvl4b8t4pkfk1364rlhbxtqhb71i2rtsckvqrvb
- kjzl6kcym7w8y5u9ktpponjvo0jh0pjl54tc9tx0pl3ayupzxaamdb7vpyduc0y

Model: kjzl6hvfrbw6c65z7bpyak6mgz9maadbucil8sey0lkuy9ebiinbjl0vd9ygal4

_**Let me know if/what additional testing is needed**_

## Definition of Done

- [x] The work addresses the description and outcomes in the issue
- [ ] I have added relevant tests for new or updated functionality
- [x] My code follows conventions, is well commented, and easy to understand
- [x] My code builds and tests pass without any errors or warnings
- [x] I have tagged the relevant reviewers
- [ ] I have updated the READMEs of affected packages
- [ ] I have made corresponding changes to the documentation
- [x] The changes have been communicated to interested parties

### DIFF
--- a/packages/graphql-scalars/src/datetime.ts
+++ b/packages/graphql-scalars/src/datetime.ts
@@ -66,7 +66,7 @@ export const DateTime = new GraphQLScalarType({
   name: 'DateTime',
   description:
     'A date-time string at UTC, such as 2007-12-03T10:15:30Z, ' +
-    'compliant with the date-time format outlined in section 5.6 of ' +
+    'compliant with the `date-time` format outlined in section 5.6 of ' +
     'the RFC 3339 profile of the ISO 8601 standard for representation ' +
     'of dates and times using the Gregorian calendar.',
   serialize: (dateString) => {
--- a/packages/graphql-scalars/src/datetime.ts
+++ b/packages/graphql-scalars/src/datetime.ts
@@ -1,0 +1,100 @@
+//@TODO Remove this file when https://github.com/Urigo/graphql-scalars/pull/1641 is merged
+
+import { GraphQLScalarType, Kind } from 'graphql'
+
+function validateTime(time: string): boolean {
+  time = time?.toUpperCase()
+  const TIME_REGEX =
+    /^([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])(\.\d{1,})?(([Z])|([+|-]([01][0-9]|2[0-3]):[0-5][0-9]))$/
+  return TIME_REGEX.test(time)
+}
+
+function validateDate(date: string): boolean {
+  const RFC_3339_REGEX = /^(\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01]))$/
+  if (!RFC_3339_REGEX.test(date)) {
+    return false
+  }
+  // Verify the correct number of days for
+  // the month contained in the date-string.
+  const year = Number(date.substr(0, 4))
+  const month = Number(date.substr(5, 2))
+  const day = Number(date.substr(8, 2))
+  switch (month) {
+    case 2: // February
+      if (((year % 4 === 0 && year % 100 !== 0) || year % 400 === 0) && day > 29) {
+        return false
+      } else if (((year % 4 === 0 && year % 100 !== 0) || year % 400 === 0) && day > 28) {
+        return false
+      }
+      return true
+    case 4: // April
+    case 6: // June
+    case 9: // September
+    case 11: // November
+      if (day > 30) {
+        return false
+      }
+      break
+  }
+  return true
+}
+
+function validateDateTime(input: string): boolean {
+  input = input?.toUpperCase()
+  const RFC_3339_REGEX =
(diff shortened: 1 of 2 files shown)


packages/graphql-scalars/src/datetime.ts
```diff
@@ -0,0 +1,100 @@
...
+  parseLiteral: (ast) => {
+    if (ast.kind !== Kind.STRING) {
+      throw new TypeError('DateTime cannot represent non string or Date type')
+    }
+    const { value } = ast
+    if (validateDateTime(value)) {
+      return value
+    }
+    throw new TypeError(`DateTime cannot represent an invalid date-time-string ${String(value)}.`)
+  },
```
PaulLeCam, line 99: Can you add the `extensions` object from https://github.com/Urigo/graphql-scalars/blob/master/src/scalars/iso-date/DateTime.ts#L70-L76 as well please to ensure the JSON schema is available?

packages/graphql-scalars/src/datetime.ts
```diff
@@ -0,0 +1,100 @@
...
+    /^(\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60))(\.\d{1,})?(([Z])|([+|-]([01][0-9]|2[0-3]):[0-5][0-9]))$/
+
+  // Validate the structure of the date-string
+  if (!RFC_3339_REGEX.test(input)) {
+    return false
+  }
+
+  // Check if it is a correct date using the javascript Date parse() method.
+  const time = Date.parse(input)
+  if (time !== time) {
```
PaulLeCam, line 54: This seems incorrect, this condition will always be `false`

packages/graphql-scalars/src/datetime.ts
```diff
@@ -0,0 +1,100 @@
...
+    'A date-time string at UTC, such as 2007-12-03T10:15:30Z, ' +
+    'compliant with the date-time format outlined in section 5.6 of ' +
+    'the RFC 3339 profile of the ISO 8601 standard for representation ' +
+    'of dates and times using the Gregorian calendar.',
+  serialize: (dateString) => {
+    if (typeof dateString === 'string' && validateDateTime(dateString)) {
+      if (dateString.indexOf('Z') !== -1) {
+        return dateString
+      } else
+        throw new TypeError(`DateTime string must be formatted to UTC time  ${String(dateString)}.`)
```
PaulLeCam, line 78: There are 2 spaces before `${String(dateString)}`, can you remove one please?

packages/graphql-scalars/src/datetime.ts
```diff
@@ -0,0 +1,100 @@
...
+    } else
+      throw new TypeError(
+        `DateTime cannot represent an invalid date-time-string  ${String(dateString)}.`
```
PaulLeCam, line 81: There are 2 spaces before `${String(dateString)}`, can you remove one please?

PaulLeCam (changes requested):
Thanks for your PR!
Looks good to me overall, just a few comments.

packages/graphql-scalars/src/datetime.ts
morozj01: I think it is supposed to check for NaN but totally agree it should be more explicit

PaulLeCam (approved):
Thanks for the changes!

PaulLeCam:
@morozj01 could you update the scalar description to use the one from graphql-scalars so it matches the existing test snapshots please?

morozj01:
> @morozj01 could you update the scalar description to use the one from graphql-scalars so it matches the existing test snapshots please?

For sure - just updated

PaulLeCam (approved):
Thanks!

PR #68, opened 2023-01-06
Add a runnable example to client setup docs

v-stickykeys:
### Summary
This PR adds npm init instructions and example code that can be pasted and run so devs can have an easy starting point to understand how to execute queries for the simple profile model definition they just created.

### Motivation
As I was going through the docs I could pretty much copy paste until this page.

### DIFF
--- a/website/versioned_docs/version-0.3.x/client-setup.mdx
+++ b/website/versioned_docs/version-0.3.x/client-setup.mdx
@@ -49,7 +49,9 @@ composedb composite:compile my-first-composite.json src/__generated__/definition
 
 Once the composite definition file is written, for example in `src/__generated__/definition.js` using the example above, it can be imported in the app to configure the client:
 
-Make sure you have the `composedb` packages [installed](installation.mdx), before running the code below.app to configure the client.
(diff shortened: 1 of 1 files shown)


pawartur:
LGTW, thanks for doing this, @v-stickykeys !! @3avi , can you also have a look? Does anything in `website/docs` should be changed too :) ?

website/versioned_docs/version-0.3.x/client-setup.mdx
```diff
@@ -65,14 +67,36 @@ More details: [`ComposeClient`](./api/classes/client.ComposeClient.md)
...
+          node {
+            displayName
+          }
+        }
+      }
     }
+  `)
+
+  for (let node of result.data.simpleProfileIndex.edges) {
+    const properties = JSON.parse(JSON.stringify(node))
```
PaulLeCam, line 87: Why do you use JSON parse/stringify here please? The data should already be JS objects that can be logged, no?

JulissaDantes:
The document this PR intended to update is no longer in our repository. Users now have code [examples](https://developers.ceramic.network/docs/composedb/guides/composedb-client/javascript-client#executing-queries) in our docs to run the queries.

PR #69, opened 2023-01-10
Fix yarn installation command for the CLI

PaulLeCam:
Reported in https://forum.ceramic.network/t/errors-with-composedb-startup/696/3


pawartur (approved):
Thanks, @PaulLeCam !

This is for yarn 1.x, right? 

FYI, I'm pretty sure that the previous command worked with yarn 2+, but it doesn't now, not sure what happened there.

PaulLeCam:
I only tried with v1, for 2+ it seems the command got removed in favor of `yarn dlx` (cf https://yarnpkg.com/getting-started/migration/#use-yarn-dlx-instead-of-yarn-global)

PR #72, opened 2023-01-25
fix: fix relay snippet not using the right variable

pradel:
## Description

fix relay snippet not using the right variable


PaulLeCam (approved):
Thanks!

PR #82, opened 2023-02-14
Updated @ceramicnetwork/* deps

pawartur:
# Updated @ceramicnetwork/* deps 

## Description

Updated to the latest RCs so that we can test the new model versioning

Also updated packages:
- @composedb/cli@0.3.2-rc.0
- @composedb/client@0.3.2-rc.0
- @composedb/devtools-node@0.3.2-rc.0
- @composedb/devtools@0.3.2-rc.0
- @composedb/graphql-scalars@0.3.2-rc.0
- @composedb/jest-environment-composedb@0.3.2-rc.0
- @composedb/runtime@0.3.2-rc.0
- @composedb/server@0.3.2-rc.0
- @composedb/types@0.3.2-rc.0

## How Has This Been Tested?

- [X] All tests pass


packages/devtools/src/schema/compiler.ts
```diff
@@ -323,6 +324,7 @@ export class SchemaCompiler {
     return {
       action: 'create',
       model: {
+        version: Composite.VERSION,
```
stbrody, line 327: hmm... I feel like calling this Composite Version isn't really exactly right.  It's still the Model version, but the view of it that's being used by the **composedb** (as opposed to "composite") tooling.  Can we call it COMPOSEDB_MODEL_VERSION or something similar?

stbrody: yeah, reading through a bit of https://github.com/ceramicstudio/js-composedb/blob/main/packages/devtools/src/composite.ts, I *think* @PaulLeCam intended Compose.VERSION to be a separate concept - a version of the Composite encoding format rather than of the Models within the Composite.  I can't be certain, but I'd still lean towards creating a new constant for this.

stbrody: Or we can just merge this as-is and have @PaulLeCam clean it up when he's back

stbrody (dismissed):
lgtm, one question around naming

packages/devtools/src/schema/compiler.ts
pawartur: Yeah, I'll add a TODO and let's merge as is and wait for @PaulLeCam to figure out what it should be. Thanks for pointing this out, @stbrody !

stbrody (approved):
lgtm

smrz2001 (approved):

packages/devtools/src/schema/compiler.ts
PaulLeCam: Yes as @stbrody commented `Composite.VERSION` represents the version of the composite structure, not individual models, that's what `Model.VERSION` can be used for.
Eventually, I think the devtools should infer the necessary model version based on the functionalities used, but as we only have a single version at the model, we can simply use `Model.VERSION`.

stbrody: The thing I was concerned about with just using Model.VERSION is what if the composedb packages update their dependency on the `@ceramicnetwork/stream-model` package and pick up a new Model.VERSION, but do no actual work to adjust to the new model format?  It seems like if there's an actual update to Model.VERSION, we want the composedb packages to break if they update without doing any thinking of what it takes to support the new version.  That's why I was thinking the composedb packages should maintain their own version of Model.VERSION that's specific to the composedb packages and needs to be updated explicitly whenever the js-ceramic side Model.VERSION is updated.  WDYT @PaulLeCam?

PaulLeCam: Yes good point, should be safer to hardcode it as `1.0` in the devtools for now

PR #91, opened 2023-02-23
Fix/skip CI tests

PaulLeCam:
I could fix an issue with the runtime package tests suite but the Ceramic node used by the CLI fails to start, might be related to https://github.com/ceramicnetwork/js-ceramic/pull/2742
Skipping the CLI tests in CI for now so at least the other tests can run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,8 @@
     "eslint-plugin-jest": "^27.2.1",
     "jest": "^29.4.2",
     "prettier": "^2.8.4",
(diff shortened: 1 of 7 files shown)


pawartur (approved):

oed (approved):
@PaulLeCam thanks, can you make sure that we have an issue tracking fixing this test suite?

PaulLeCam:
Yes, I'll add a ticket to track down and fix the issue

PR #95, opened 2023-02-26
Include instructions to list models on clay

gvelez17:
# Ensure hackathon devs can find models 

## Description

By default we list models on mainnet, but there are not any models currently on mainnet

Tell new developers how to find models on clay as well

Tested manually.


3avi (approved):
includes how to find models on clay

PR #96, opened 2023-02-27
Clarify that only clay nodes that index the set of models work for model:list

stbrody:
Follow-up to https://github.com/ceramicstudio/js-composedb/pull/95/ to clarify that not just any clay node will work


stbrody:
Note I still don't like broadcasting this node URL and hope we can take this instruction out of our docs and replace it with something more straightforward for our users that also lets us obfuscate that node a little bit more.

3avi (approved):
noted spencer - this is temporary

PR #98, opened 2023-02-27
Update example to use new --network flag to get models from clay

stbrody:
Depends on https://github.com/ceramicstudio/js-composedb/pull/97 being merged


PaulLeCam (approved):
Thanks, let's just wait until the new CLI version is released to update the docs

PR #108, opened 2023-03-01
Add another long relation example

dbcfd:


3avi (approved):
Good stuff. One of our grants cohort folks saw the example already btw!

PR #115, opened 2023-04-21
Running in the cloud k8s do

3benbox:
# [Running in the cloud k8s do] - #PLAT-1393

## Description

Steps to setup a kubernetes cluster with Digital Ocean and deploy the ceramic stack.

## How Has This Been Tested?

Looking for testers!

PR #117, opened 2023-04-26
Updated infra script guide

JustinaPetr:
# [Adoption k8s infra script] - #ADOPT-1199

## Description

This is an updated version of Ben's work: https://github.com/ceramicstudio/js-composedb/pull/115 
I opened a new PR because there were quite a few cosmetic changes that I added, also slightly edited the original guide.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,555 +1,401 @@
-lockfileVersion: '6.0'
+lockfileVersion: 5.4
 
 importers:
 
   .:
+    specifiers:
+      '@changesets/cli': ^2.26.1
+      '@jest/globals': ^29.5.0
+      '@skypack/package-check': ^0.2.2
+      '@swc/cli': ^0.1.62
+      '@swc/core': ^1.3.52
+      '@swc/jest': ^0.2.26
+      '@types/jest': ^29.5.1
+      '@typescript-eslint/eslint-plugin': ^5.59.0
+      '@typescript-eslint/parser': ^5.59.0
+      del-cli: ^5.0.0
+      eslint: ^8.38.0
+      eslint-config-3box: ^0.4.1
+      eslint-plugin-jest: ^27.2.1
+      jest: ^29.5.0
+      prettier: ^2.8.7
+      turbo: ^1.9.3
+      typedoc: 0.24.4
+      typedoc-plugin-markdown: ^3.15.1
+      typescript: ^5.0.4
     devDependencies:
-      '@changesets/cli':
-        specifier: ^2.26.1
-        version: 2.26.1
-      '@jest/globals':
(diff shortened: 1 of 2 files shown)


PaulLeCam:
I went through the Kubernetes setup on Digital Ocean, here are my notes:

## Specific issues/questions

### Creating a Kubernetes Cluster

> When it comes to choosing your cluster capacity, we recommend starting with the smallest cluster size and upgrading later. For example, to follow this guide you can start with a 1GB RAM and 1vCPU cluster.

Is there anything smaller than Basic - 2vCPU 4 GB RAM available on Digital Ocean?

### Access the ComposeDB API

Could be useful to separate commands from outputs so they can directly copied, ex:

```sh
kubectl port-forward --namespace ceramic composedb-0 7007:7007

Forwarding from 127.0.0.1:7007 -> 7007
Forwarding from [::1]:7007 -> 7007
```

Could instead be:

Command

```sh
kubectl port-forward --namespace ceramic composedb-0 7007:7007
```

Expected output

```sh
Forwarding from 127.0.0.1:7007 -> 7007
Forwarding from [::1]:7007 -> 7007
```

### What is my admin DID?

Why is the admin DID seed encoded in base 64 where we use base 16 in the CLIs please?

Also, I couldn't get access to the seed when running `kubectl get secrets  ceramic-admin -o yaml`, I got the following error: `Error from server (NotFound): secrets "ceramic-admin" not found`.

Running `kubectl get secrets` gives `No resources found in default namespace.`, though I checked I ran `./k8s/base/composedb/create-secrets.sh` in a previous step, and trying to run it again outputs:

```sh
error: failed to create secret secrets "postgres-auth" already exists
error: failed to create secret secrets "ceramic-admin" already exists
```

Any idea what I missed please?

## Overall feedback

The tutorial worked well for deploying to Digital Ocean up to the point of calling the healthcheck API, but then I couldn't access my admin DID.
Overall the tutorial is very clear, the most confusing thing I ran into is the switch between local/generic sections and Digital Ocean specific ones, notably reaching `kubectl port-forward --namespace ceramic composedb-0 7007:7007` that is the local command. Maybe the Digital Ocean steps could be more linear/separated from the generic ones?

JustinaPetr:
Thank you for great feedback @PaulLeCam. I made some cosmetic updates following your recommendation to separate commands and the outputs and added a few clarifications to make sure people can follow the steps easily.

Regarding the `kubectl get secrets` error, looks like that command needed a namespace specification. Just added it there. 

Will think about how we can separate the DO specific commands and local commands more.

JustinaPetr:
@3avi Can I get a review here? If it looks good, we could potentially merge it so that we can go ahead with testing.

3avi (approved):
lgtm

PR #122, opened 2023-05-10
feat(infra-scripts): Add guide for utilizing composedb with the deployed assets

dbcfd:


3avi (approved):
thanks for making this work with ComposeDB and Graphiql, LGTM

PR #128, opened 2023-05-25
CDB-2492: Add support for parsing index directive

dbcfd:


packages/devtools/src/schema/typeDefinitions.ts
```diff
@@ -49,4 +49,13 @@ directive @createModel(
 ) on OBJECT
 
 directive @loadModel(id: StreamID!) on OBJECT
+
+input IndexField {
+  path: [String!]!
+}
+
+directive @createIndex(
```
PaulLeCam, line 57: I still think we should define indices as an argument on the `@loadModel` and `@createModel` directives to make it explicit indices are associated to models, for example this should be invalid:

```graphql
type Image @createIndices(...) {
  src: String!
  alt: String
}

type Post @createModel(...) {
  cover: Image!
  ...
}
```

Instead the association would be more explicit with something like:

```graphql
type Image {
  src: String!
  alt: String
}

type Post @createModel(..., indices: [{fields: [{path: ["cover", "src"]}]}]) {
  cover: Image!
  ...
}
```

packages/devtools/src/schema/typeDefinitions.ts
```diff
@@ -49,4 +49,13 @@ directive @createModel(
...
+  name: String!
```
PaulLeCam, line 58: Can we make the `name` optional please?
Also, what error message gets raised on duplicate names? Is is properly caught by the tools?

packages/devtools/src/schema/parser.ts
```diff
@@ -95,9 +97,12 @@ export class SchemaParser {
         // return type
       },
       [MapperKind.OBJECT_TYPE]: (type: GraphQLObjectType) => {
+        const directives = getDirectives(this.#schema, type)
+        const indices = this._parseIndices(directives)
```
PaulLeCam, line 101: Indices can only be set on models so I think the logic should be different here, only trying to parse the indices after having identified the object is a model and throwing an error if the `@createIndex` directive is used on embedded objects.

PaulLeCam (changes requested):

packages/devtools/src/composite.ts
```diff
@@ -550,16 +551,14 @@ export class Composite {
...
-      definedIndices = this.#definition.indices.models
+    let definedIndices: Array<ModelFieldsIndex> = []
+    if (this.#definition.indices && this.#definition.indices.indices) {
+      definedIndices = this.#definition.indices.indices
     }
-    Object.keys(this.#definition.models).forEach((key) => {
-      const indices = definedIndices[key]
-      modelIndices.set(key.toString(), indices)
+    const modelIds = Array.from(Object.keys(this.#definition.models)).map((id) => {
+      return StreamID.fromString(id)
```
PaulLeCam, line 559: Why are the model IDs parsed from string here please? They'll need to be converted to string in the API call.

packages/devtools/test/schema.test.ts
```diff
@@ -757,6 +757,14 @@ describe('schema', () => {
         fields: [{
           path: ["idValue"]
         }],
+      )
+      @createIndex(
+        name: "id_required_index",
```
PaulLeCam, line 762: Can you add a test case for the error thrown when the index name already exists please?

PaulLeCam (changes requested):

packages/devtools/src/composite.ts
dbcfd: The http API deals in json, and thus uses strings. The API used by composedb is typescript, and uses StreamID. RemoteAdminAPI may use http underneath, but we shouldn't leak transport details to the user of the API.

PaulLeCam: But that's an unnecessary change here. The client and APIs already work with strings, there no need to make such string (client) -> StreamID (here) -> string (transport) -> StreamID (server) conversions where we could just be string (client) -> string (transport - no change) -> StreamID (server). I expect the server also needs the models IDs as string anyway to interact with the DB, so really the conversion to StreamID doesn't bring anything to the flow.

PaulLeCam: OK I guess we're doing these kinds of unnecessary conversions all over the place already so one more isn't going to hurt.

packages/types/src/index.ts
```diff
@@ -63,6 +65,10 @@ export type CompositeDefinitionType<T> = {
    * Optional composite-level views.
    */
   views?: CompositeViewsDefinition
+  /**
+   * Optional composite-level indices
+   */
+  indices?: Map<string, Array<FieldsIndex>>
```
PaulLeCam, line 71: Should be a `Record` rather than `Map` to match the JSON serialization types

packages/devtools/src/schema/parser.ts
```diff
@@ -628,6 +659,12 @@ export class SchemaParser {
   }
 }
 
+// function parseStringValue(arg: GraphQLArgument): string {
```
PaulLeCam, line 662: To remove if not needed please

PaulLeCam (changes requested):

packages/devtools/src/schema/parser.ts
```diff
@@ -140,6 +155,18 @@ export class SchemaParser {
     return this.#def
   }
 
+  _parseIndices(directives: Array<DirectiveAnnotation>): Array<FieldsIndex> {
+    return directives.flatMap((d) => {
+      if (d.name === 'createIndex' && d.args) {
+        let indexName = undefined
+        const fields = d.args.fields as Array<Field>
+        return [{ name: indexName, fields }]
```
PaulLeCam, line 163:
```suggestion
        return [{ fields: d.args.fields as Array<Field> }]
```
The `name` is always `undefined` so this could simplified a bit

packages/devtools/package.json
```diff
@@ -39,9 +39,9 @@
     "prepublishOnly": "package-check"
   },
   "dependencies": {
-    "@ceramicnetwork/common": "^2.25.0",
-    "@ceramicnetwork/stream-model": "^1.7.0",
-    "@ceramicnetwork/streamid": "^2.14.0",
+    "@ceramicnetwork/common": "2.30.0-rc.0",
```
PaulLeCam, line 42: Why is `common` saved as exact version and not the others please? That could create issues with type mismatches.

packages/devtools/src/composite.ts
```diff
@@ -552,14 +552,13 @@ export class Composite {
   async startIndexingOn(ceramic: CeramicApi): Promise<void> {
     const definedIndices = Array.from(Object.keys(this.#definition.models)).map((id) => {
       const streamID = StreamID.fromString(id)
-      const indices = this.#definition.indices?[id]
+      const indices = (this.#definition.indices ?? {})[id]
```
PaulLeCam:
```suggestion
      const indices = this.#definition.indices?.[id]
```

packages/jest-environment-composedb/package.json
```diff
@@ -23,7 +23,7 @@
     "lint": "eslint index.js --fix"
   },
   "dependencies": {
-    "@ceramicnetwork/core": "^2.31.0",
+    "@ceramicnetwork/core": "2.36.0-rc.0",
```
PaulLeCam: Does it need to be this exact version?

packages/server/package.json
```diff
@@ -39,15 +39,15 @@
...
+    "@ceramicnetwork/http-client": "^2.27.0-rc.0",
     "@composedb/constants": "workspace:^0.4.0",
     "@composedb/runtime": "workspace:^0.4.2",
     "get-port": "^6.1.2",
     "graphql": "^16.5.0",
     "graphql-yoga": "^3.9.1"
   },
   "devDependencies": {
-    "@ceramicnetwork/common": "^2.24.0",
+    "@ceramicnetwork/common": "2.30.0-rc.0",
```
PaulLeCam, line 50: Exact version?

PaulLeCam (commented):
LGTM mostly wondering about the different version range constraints set in the different packages

packages/devtools/package.json
dbcfd: Since it's an RC, was doing exact versions as a way to force them to be changed once a release is out. They should all be exact version. Will make sure I've done that.

packages/devtools/src/schema/typeDefinitions.ts
dbcfd: Can revisit this after dogfooding

packages/cli/globalSetup.js
```diff
@@ -56,7 +56,7 @@ export default async function globalSetup() {
   await fs.writeJson(CONFIG_PATH, TEST_DAEMON_CONFIG)
 
   globalThis.servers = await setup({
-    command: `IPFS_PATH=\'${GOIPFS_DIR_PATH.pathname}\' pnpm dlx @ceramicnetwork/cli daemon --config ${CONFIG_PATH.pathname}`,
+    command: `IPFS_PATH=\'${GOIPFS_DIR_PATH.pathname}\' pnpm dlx @ceramicnetwork/cli@2.35.0-rc.0 daemon --config ${CONFIG_PATH.pathname}`,
```
stbrody, line 59: woah, seems weird and dangerous to hard code a specific version here like this.  At least it should probably be `@next` although even that feels a little sketch

packages/cli/package.json
```diff
@@ -69,11 +69,11 @@
     ]
   },
   "dependencies": {
-    "@ceramicnetwork/3id-did-resolver": "^2.19.0",
-    "@ceramicnetwork/http-client": "^2.23.0",
-    "@ceramicnetwork/stream-model": "^1.8.0",
-    "@ceramicnetwork/stream-model-instance": "^1.8.0",
-    "@ceramicnetwork/streamid": "^2.14.0",
+    "@ceramicnetwork/3id-did-resolver": "2.23.0-rc.0",
```
stbrody, line 72: why pinning to specific versions?  We should be able to keep the `^` here and be guaranteed to get a version at *least* after that RC, but still be able to pick up newer versions as they get released

packages/client/test/client.test.ts
```diff
@@ -33,7 +33,7 @@ describe('client', () => {
     expect(res.data?.createGenericProfile.document.id).toBeDefined()
   }, 30000)
 
-  test('create and query post with comments', async () => {
+  test.skip('create and query post with comments', async () => {
```
stbrody, line 36: why do we need to skip this test?  Can you add a comment explaining what's wrong with it and a TODO referencing a linear ticket number where it will be restored?

packages/devtools/src/composite.ts
```diff
@@ -37,7 +38,8 @@ function isSignedCommit(input: Record<string, any>): input is SignedCommit {
 function toStrictDefinition(definition: InternalCompositeDefinition): StrictCompositeDefinition {
   const emptyViews = { account: {}, root: {}, models: {} }
   const views = definition.views ? { ...emptyViews, ...definition.views } : emptyViews
-  return { aliases: {}, commonEmbeds: [], ...definition, views }
+  const indices = definition.indices ? definition.indices : {}
```
stbrody, line 41: is `indices` an object or an array?  Should the default be `[]` instead of `{}`?

packages/devtools/src/schema/typeDefinitions.ts
stbrody: I agree with leaving this as-is for now, but let's definitely not forget to revisit this.  I think experimenting (and possibly even A/B testing) with different syntax for index definitions is something we should really consider a priority during the testing and feedback phase before release - and we should be careful to resist the temptation to push this out to a full release before we've gotten proper usability feedback on the API choices we made.  CC @jpham2023 @michaelsena to put this on your radar

packages/devtools/src/schema/types.ts
```diff
@@ -4,10 +4,14 @@ import type {
   ModelRelationsDefinition,
   ModelViewsDefinition,
 } from '@ceramicnetwork/stream-model'
-import type { RuntimeViewField } from '@composedb/types'
+import type { RuntimeViewField, FieldsIndex } from '@composedb/types'
 
 import type { ScalarSchema } from '../types.js'
 
+export type Field = {
```
stbrody, line 11: is this redundant of the same-named type from js-ceramic (https://github.com/ceramicnetwork/js-ceramic/blob/8aaec9b820c95d7037eb4a84192ce7eef56072c5/packages/common/src/ceramic-api.ts#L14)?

Can we just import and use the types from js-ceramic instead of duplicating them?

packages/jest-environment-composedb/index.js
```diff
@@ -57,9 +57,25 @@ export default class ComposeEnvironment extends NodeEnvironment {
...
 
   async teardown() {
-    await this.global.ceramic.close()
-    await this.global.ipfs.stop()
-    await this.tmpFolder.cleanup()
-    await super.teardown()
+    try {
+      await this.global.ipfs.stop()
+    } catch (e) {
+      //ignore
```
stbrody, line 63: this worries me.  We should be aiming to maintain the ability to do a clean shutdown, this will just mask if we have a problem there.  Were you actually hitting errors here?

packages/cli/globalSetup.js
dbcfd: For this, we specifically want that RC, because otherwise things don't work. It's something that should be removed when we go from RC to release.

stbrody: We want *at least* that RC.  No reason we couldn't take the full release once that RC is promoted to full release

packages/cli/package.json
dbcfd: Since this is also for an RC, we want to know exactly what versions are being for things that may change, e.g. js-ceramic, so we can more easily address bugs or know if a bug has been fixed and this RC just has out of date versions.

packages/devtools/src/composite.ts
dbcfd: It's a record.

packages/jest-environment-composedb/index.js
dbcfd: Yep, was actually hitting errors here, which sadly can't be handled by this. It actually took reordering of the stop.

packages/cli/package.json
stbrody: I don't understand why we wouldn't want to automatically pick up the fully released version once the RC is promoted to full release

packages/jest-environment-composedb/index.js
stbrody: What are the errors?  Why are we only seeing them here now?  Is there a regression of some kind?  At the very least we should file a ticket and add a TODO here to remove the try/catches once the shutdown issue is fixed, if we can't just go ahead and fix the shutdown issue directly.

packages/client/test/client.test.ts
dbcfd: No longer skipped

packages/devtools/src/schema/types.ts
dbcfd: exported from js-ceramic now

packages/jest-environment-composedb/index.js
dbcfd: The main one that occurs is `aborted` on sqlite. As ceramic is in the process of closing, it receives additional events it attempts to write to sqlite (maybe from IPFS?), which then causes an error, and the test fails. Given that this is a test, this feels like the correct behavior, since it shouldn't be up to this library's tests to determine if ipfs and ceramic are closing properly.

packages/cli/globalSetup.js
dbcfd: For this RC we want _exactly that_ RC. That way if a new RC is released for ceramic, we can validate it, before we push a new RC of js-composedb. This is the same logic as the dependencies. An RC for ceramic could come out that the RC of js-composedb cannot automatically handle. All RC dependencies in js-composedb are adjusted once we determine they should be pulled into js-composedb.

packages/jest-environment-composedb/index.js
dbcfd: https://linear.app/3boxlabs/issue/CDB-2615/determine-why-shutting-down-ceramic-before-ipfs-causes-aborted

packages/cli/globalSetup.js
dbcfd: https://www.notion.so/threebox/Query-Filters-Query-Ordering-01e9acb5d8b547b89d0ee43dacca3007?pvs=4#0f9ad767d5234fc7bef9c0af01cc1642

stbrody: That is a pretty major change in the way we have always managed the dependency between js-ceramic and js-composedb.  It also seems very backwards to me.  If we do a new RC of js-ceramic, it's most likely because of something we found in testing and is likely something we would want to pick up and include in the next release of js-composedb.  Pinning ourselves to an older RC just adds friction to our testing process and makes it take longer to start getting feedback on if our upcoming release of js-ceramic has problems that would break js-composedb.  If the new RC of js-ceramic is broken we want to learn that ASAP.

willex (approved):
LGTM

packages/jest-environment-composedb/index.js
PaulLeCam: I don't think it's a good idea to merge changes breaking the test setup and have a "let's fix it later" ticket in the backlog.
If the Ceramic RC fails to properly close its processes, that seems like a high priority issue that should be addressed IMO.

dbcfd: That should be addressed in js-ceramic, but should not composedb RC's. At worst, this should be a known issues for this RC.

packages/cli/globalSetup.js
dbcfd: added https://linear.app/3boxlabs/issue/CDB-2618/determine-appropriate-strategy-for-using-js-ceramic-rcs-in-js

packages/cli/package.json
dbcfd: https://linear.app/3boxlabs/issue/CDB-2618/determine-appropriate-strategy-for-using-js-ceramic-rcs-in-js

PR #130, opened 2023-06-01
Check for authenticated DID when creating or indexing models

PaulLeCam:


stbrody (commented):
Should we include any guidance on how to fix the error to users?  If this error is only ever going to come from the CLI tool, should we say the name of the CLI flag that needs to be provided?

PaulLeCam:
The logic isn't specific to the CLI, the devtools can be used in scripts as well.
The ComposeDB docs used to be explicit about the fact the key needs to be provided for some commands but it seems this has changed in the new versions so that might be a better way to address the issue.

PR #131, opened 2023-06-02
Feat/add pkh suport for deploy composite 

qbig:
# Add did:pkh Support for create composite models - #[Issue]

## Description

https://forum.ceramic.network/t/only-admin-could-create-models-composites/1135/22

## How Has This Been Tested?

- [ ] Test A (e.g. Test A - New test that ... ran in local, docker, and dev unstable.)
- [ ] Test B

## Definition of Done

- [ ] The work addresses the description and outcomes in the issue
- [ ] I have added relevant tests for new or updated functionality
- [ ] My code follows conventions, is well commented, and easy to understand
- [ ] My code builds and tests pass without any errors or warnings
- [ ] I have tagged the relevant reviewers
- [ ] I have updated the READMEs of affected packages
- [ ] I have made corresponding changes to the documentation
- [ ] The changes have been communicated to interested parties

## References:

Please list relevant documentation (e.g. tech specs, articles, related work etc.) relevant to this change, and note if the documentation has been updated.


oed (changes requested):
We should not need a network call for this

packages/devtools/src/composite.ts
```diff
@@ -57,14 +65,98 @@ function assertModelsHaveCommits(
...
+      )
+    }
+  } else if (!model.metadata.controller.startsWith('did:key:')) {
+    throw new Error(
+      `${unsupported}, only did:key is supported`
+    )
+  }
+}
+
+async function assertSupportedReadModelController(model: Model, signedCommit: SignedCommit): Promise<void> {
```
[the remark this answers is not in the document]
oed: @PaulLeCam Looks like we got some type signatures incorrect?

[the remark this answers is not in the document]
qbig: [SignedCommitContainer](https://github.com/ceramicnetwork/js-ceramic/blob/03b544717b295eb5134a890720c4b15e278f6ca1/packages/common/src/stream.ts#LL63C1-L63C1) was cast to SignedCommit via here `commit as unknown as SignedCommit`
```typescript
      const [genesis, ...updates] = commits as Array<StreamCommits>
      const model = await ceramic.createStreamFromGenesis<Model>(
        Model.STREAM_TYPE_ID,
        genesis,
        MODEL_GENESIS_OPTS
      )
      for (const commit of updates) {
        const signedCommit = commit as unknown as SignedCommit
        assertSupportedReadModelController(model, signedCommit)
        await ceramic.applyCommit(model.id, signedCommit)
      }
```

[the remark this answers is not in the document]
PaulLeCam: Yes definitely possible these types are incorrect, don't hesitate to fix in this PR

packages/devtools/src/composite.ts
```diff
@@ -17,6 +19,12 @@ import { createRuntimeDefinition } from './formats/runtime.js'
 import { createAbstractCompositeDefinition } from './schema/compiler.js'
 import type { AbstractModelDefinition } from './schema/types.js'
 
+// The hardcoded "model" StreamID that all Model streams have in their metadata. This provides
+// a "model" StreamID that can be indexed to query the set of all published Models.
+// The StreamID uses the "UNLOADABLE" StreamType, and has string representation: "kh4q0ozorrgaq2mezktnrmdwleo1d"
+
+const MODEL_STREAM_ID = 'kh4q0ozorrgaq2mezktnrmdwleo1d'
```
PaulLeCam, line 26:
```suggestion
const MODEL_STREAM_ID = Model.MODEL.toString()
```
Better use the stream ID defined on the model itself than copying it I think.

packages/devtools/src/composite.ts
```diff
@@ -57,14 +65,98 @@ function assertModelsHaveCommits(
...
+      )
+    }
+    const cacao: Cacao = ceramic.context.did.capability(); 
+    if (cacao.p.exp == null) {
+      throw new Error(
+        `${unsupported}, only did:pkh CACAO without expiry is supported`
+      )
+    }
+    const hasModelResource = cacao.p.resources?.includes(`ceramic://*?model=${MODEL_STREAM_ID}`)
+    if (cacao.p.resources?.length != 1 || !hasModelResource) {
```
PaulLeCam, line 88:
```suggestion
    if (!hasModelResource) {
```
Seems incorrect to ensure only one resource is allowed in the CACAO, no? The CACAO could be used for other resources as well.

packages/devtools/src/composite.ts
```diff
@@ -57,14 +65,98 @@ function assertModelsHaveCommits(
...
+      throw new Error(
+        `${unsupported}, only cacao with resource ${MODEL_STREAM_ID} is supported`
+      )
+    }
+  } else if (!model.metadata.controller.startsWith('did:key:')) {
+    throw new Error(
+      `${unsupported}, only did:key is supported`
```
PaulLeCam, line 95:
```suggestion
      `${unsupported}, only did:key and did:pkh are supported`
```

PaulLeCam (changes requested):
Thanks for your PR, looks good overall but I wonder if this logic should live on the Model handler implementation rather than here?
It would be safer to ensure all models are validated rather than only the ones read or written using our devtools, what do you think @oed please?

oed:
> I wonder if this logic should live on the Model handler implementation rather than here?

@PaulLeCam I think this would only be a minor improvement since any streamtype specific logic won't be enforced on the protocol level after the migration to the rust implementation. I think we could add a check there, but I think logic here will be more impactful.

packages/devtools/src/composite.ts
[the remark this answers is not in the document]
oed: @qbig If you console.log the commit I think you will get a [SignedCommitContainer](https://github.com/ceramicnetwork/js-ceramic/blob/03b544717b295eb5134a890720c4b15e278f6ca1/packages/common/src/stream.ts#LL63C1-L63C1)?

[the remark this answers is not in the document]
oed: Yep, SignedCommitContainer = DagJWSResult.

So you can decode the `cacaoBlock` (which will be present when did:pkh is used) using dag-cbor to get the Cacao object.

packages/devtools/src/composite.ts
```diff
@@ -63,10 +70,68 @@ function assertModelsHaveCommits(
...
+        `${unsupported}, did missing from ceramic context`
+      )
+    }
+    if (!ceramic.context.did.hasCapability()) {
+      throw new Error(
+        `${unsupported}, only did:pkh with CACAO is supported`
+      )
+    }
+    const cacao: Cacao = ceramic.context.did.capability(); 
+    if (cacao.p.exp == null) {
```
oed, line 87: I think this logic is backwards? We should only allow creation if ocap doesn't have expiry?
```suggestion
    if (cacao.p.exp !== null) {
```

packages/devtools/src/composite.ts
```diff
@@ -63,10 +70,68 @@ function assertModelsHaveCommits(
...
+  }
+  if (cacao.p.exp == null) {
+    throw new Error(
+      `only did:pkh CACAO without expiry is supported`
+    )
+  }
+  const hasModelResource = cacao.p.resources?.includes(`ceramic://*?model=${MODEL_STREAM_ID}`)
+  if (!hasModelResource) {
+    throw new Error(
+      `only cacao with resource ${MODEL_STREAM_ID} is supported`
```
oed, line 134:
```suggestion
      `only cacao with resource "ceramic://*?model=${MODEL_STREAM_ID}" is supported`
```
We can make this error more explicit!

packages/devtools/src/composite.ts
```diff
@@ -63,10 +70,68 @@ function assertModelsHaveCommits(
...
+      `${unsupported}, only did:key is supported`
+    )
+  }
+}
+
+function assertValidCacao(cacao: Cacao, controller: string): void {
+  if (cacao.p.iss != controller) {
+    throw new Error(`Cacao issuer ${cacao.p.iss} does not match controller ${controller}`)
+  }
+  if (cacao.p.exp == null) {
```
oed, line 126:
```suggestion
  if (cacao.p.exp !== null) {
```
This one also looks incorrect?

packages/devtools/src/composite.ts
```diff
@@ -116,9 +181,10 @@ async function loadModelsFromCommits<Models = Record<string, StreamCommits>>(
         genesis,
         MODEL_GENESIS_OPTS
       )
-      assertSupportedModelController(model)
       for (const commit of updates) {
-        await ceramic.applyCommit(model.id, commit as unknown as SignedCommit)
+        const signedCommitContainer = commit as unknown as SignedCommitContainer
+        assertSupportedReadModelController(model, signedCommitContainer)
```
oed, line 186: Shouldn't we only do this check if it's a signed container? AnchorCommits won't be signed.

oed: We also need to check if `genesis` with `assertSupportedReadModelController`.

oed (changes requested):
This looks much better. A few updates needed.

packages/devtools/src/composite.ts
qbig: > Shouldn't we only do this check if it's a signed container? AnchorCommits won't be signed.

Do we need to validate AnchorCommit here? possible to do it without network request?

qbig: > Shouldn't we only do this check if it's a signed container? AnchorCommits won't be signed.
I was assuming they are all signed commit from the function signature. How do I handle the anchor commit here ?
```typescript
async function loadModelsFromCommits<Models = Record<string, StreamCommits>>(
  ceramic: CeramicApi,
  modelsCommits: Models & {} // eslint-disable-line @typescript-eslint/ban-types
): Promise<Record<keyof Models, ModelDefinition>> {
//...
export type StreamCommits = Array<DagJWSResult>;
//...
export declare type SignedCommitContainer = DagJWSResult;

```

qbig: > We also need to check if `genesis` with `assertSupportedReadModelController`.

How about `static async fromModels(params: FromModelsParams): Promise<Composite> ` Do we also need to validate AnchorCommit and genesis commit? Does `loadStreamCommits` only return genesis and signed commits?

PaulLeCam: `loadStreamCommits()` loads all the commits, including anchor ones, but they are filters to only signed commits in the following logic.
Assuming there is no expiration date on the CACAO I don't think we need to care about the anchor commits, unless I'm missing something? cc @oed

oed: We don't need to validate the anchor commits.

oed (dismissed):
lgtm, would be good to get another review from @PaulLeCam as well.

packages/devtools/src/composite.ts
```diff
@@ -63,10 +70,68 @@ function assertModelsHaveCommits(
...
+      `${unsupported}, only did:key is supported`
+    )
+  }
+}
+
+function assertValidCacao(cacao: Cacao, controller: string): void {
+  if (cacao.p.iss != controller) {
+    throw new Error(`Cacao issuer ${cacao.p.iss} does not match controller ${controller}`)
+  }
+  if (cacao.p.exp !== null) {
```
PaulLeCam, line 126:
```suggestion
  if (cacao.p.exp != null) {
```

PaulLeCam (changes requested):
Thanks for the changes, mostly LGTM but I just spotted a few things.

oed (dismissed):
I think this should be good to go. Will let @PaulLeCam have another look before we merge.

PaulLeCam (approved):
Thanks for the changes!

PaulLeCam:
@qbig looks like there is an issue with the lockfile, can you make sure it is updated please?

PaulLeCam:
@qbig could you make sure your branch is in sync with the `main` branch and passes the checks locally (running `pnpm install`, `pnpm run lint` and `pnpm test`) please?

PaulLeCam (approved):

PaulLeCam:
Thanks @qbig!

PR #135, opened 2023-06-27
Query filters and sorting inputs

PaulLeCam:
Will need to be updated to only generate inputs for indexed fields once https://github.com/ceramicstudio/js-composedb/pull/128 is merged + use stable versions of Ceramic packages once released but the logic won't change much so should be OK to review already.


dbcfd:
closing for https://github.com/ceramicstudio/js-composedb/pull/144

PR #136, opened 2023-07-10
v0.5 preview docs setup

PaulLeCam:
Like we did before the v0.4 release, this creates a new "preview" version in the docs for changes that should become v0.5.


JustinaPetr (approved):
Thank you for adding all this. These changes look good to me. Examples in the "Guides" section will be very useful here.

PR #144, opened 2023-07-17
CDB-2531 - Generate Query Inputs

dbcfd:


stbrody (commented):
can we set the base branch for this PR to the branch from https://github.com/ceramicstudio/js-composedb/pull/128 so reviewers can see just the new changes on top of that other PR?

dbcfd:
> can we set the base branch for this PR to the branch from #128 so reviewers can see just the new changes on top of that other PR?

@stbrody done

willex (approved):
Looks good! Looking forward to dogfooding on this.

packages/devtools/src/formats/runtime.ts
```diff
@@ -249,6 +251,35 @@ export class RuntimeModelBuilder {
   }
 }
 
+function applyIndexing(
```
stbrody, line 254: this function is mutating its inputs which is a bit non-obvious if you don't read carefully.  Can you please add some doccomments explaining what this function is doing?

packages/runtime/src/query.ts
```diff
@@ -1,27 +1,19 @@
...
   PaginationQuery,
+  QueryFilters,
   StreamState,
 } from '@ceramicnetwork/common'
 import { ModelInstanceDocument } from '@ceramicnetwork/stream-model-instance'
 import type { Connection, ConnectionArguments } from 'graphql-relay'
 
 export type ConnectionQuery = BaseQuery & ConnectionArguments
 
-export function toBaseQuery(source: BaseQuery): BaseQuery {
```
stbrody, line 14: I guess the point of this function was to strip out any extra params that aren't expected?  Is there a reason why you've removed it now?  What changed that this is no longer the desired behavior?

packages/runtime/src/schema.ts
```diff
@@ -451,44 +556,66 @@ class SchemaBuilder {
...
+      case 'queryCount': {
+        const filtersObj = this.#inputObjects[`${modelAlias}Filters`]
+        const args: ObjMap<GraphQLArgumentConfig> = {
+          account: {
+            type: GraphQLID,
+            description: 'Counts only documents created by the provided account',
+          },
+        }
+        if (filtersObj) {
+          args.filters = { type: filtersObj }
```
stbrody, line 602: should this have a `description` field like the `account` arg does?

stbrody (approved):
this honestly feels a bit out of my depth to be able to give a proper review.  I'm going to go ahead and approve for now to unblock testing, but I'm also going to add @PaulLeCam to the PR and suggest that he do a post-merge review when he's back next week

packages/runtime/src/query.ts
dbcfd: I think this is now covered by https://github.com/ceramicstudio/js-composedb/pull/144/files#diff-e080742723a30fbf28f48db65ef61685cc18bf45ef6de97ed5df6ce7b20597acR181 where we pass query filters. 

Tagging @PaulLeCam so he can verify.

packages/runtime/src/schema.ts
dbcfd: I think @PaulLeCam added this to provide more description on this field, whereas other fields are mostly self explanatory.

willex (approved):

PR #147, opened 2023-07-21
chore: Add a workflow for releasing, candidate or GA

dbcfd:


.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          fetch-depth: 0
+      - name: setup node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+      - name: install pnpm
+        run: npm i pnpm@latest -g
+      - name: Setup npmrc
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
```
PaulLeCam, line 32: How/where is this secret defined please?

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: pnpm changeset pre enter rc
+      - name: create and publish versions
+        uses: changesets/action@v1
+        with:
+          version: pnpm ci:version
+          commit: "chore: update versions"
+          title: "chore: update versions"
+          publish: pnpm ci:publish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
PaulLeCam, line 48: Is this needed? If so, how is this token defined please?

.github/workflows/release.yml
dbcfd: This is a secret defined for the action, that can be at repository or organization level. https://docs.github.com/en/actions/security-guides/encrypted-secrets

.github/workflows/release.yml
dbcfd: This is a secret provided by github which allows actions to do actions against the repository, such as creating pull requests or pushing code. https://docs.github.com/en/actions/security-guides/automatic-token-authentication

PR #150, opened 2023-07-27
fix: CDB-2630 No Filters in Graphiql

dbcfd:
Compiled runtime composites now include indices so that filters can be generated.


willex (approved):
Awesome, thanks for the quick turnaround!

willex (approved):
Boom!

PR #152, opened 2023-07-31
fix: cdb-2632 Fix issues using other input types

dbcfd:


packages/runtime/src/schema.ts
```diff
@@ -166,12 +171,16 @@ const valueFilterInputs = {
   FloatValueFilter: createScalarValueFilterInput(GraphQLFloat),
   IntValueFilter: createScalarValueFilterInput(GraphQLInt),
   StringValueFilter: createScalarValueFilterInput(GraphQLString),
+  DateValueFilter: createScalarValueFilterInput(GraphQLDate),
+  DateTimeValueFilter: createScalarValueFilterInput(GraphQLDateTime),
```
PaulLeCam, line 175: These shouldn't be needed, scalars that extend from `String` can simply use the `StringValueFilter`

packages/runtime/test/runtime.test.ts
```diff
@@ -141,30 +141,63 @@ describe('runtime', () => {
...
         }
       }
-      `,
-    )
-    expect(res).toMatchSnapshot()
+    }`
+
+    const filterVars1 = {
+      input: {
+        and: [{ where: { ranking: { greaterThan: 4 } } }, { where: { ranking: { lessThan: 7 } } }],
```
PaulLeCam, line 180:
```suggestion
        where: { ranking: { greaterThan: 4, lessThan: 7 } },
```
This should be supported as-is.

PaulLeCam (changes requested):

packages/runtime/src/schema.ts
willex: @dbcfd any advantage using specific types here?

willex (approved):

packages/runtime/src/schema.ts
willex: @PaulLeCam followed up with Danny around this as well. It seems using those types makes query introspection work better with other clients e.g. Apollo which would come in handy because I think thats what devs are using for actual app implementation (React) and then responds raise errors correctly when non dates are being passed in. If you don't see any major issue otherwise I would vote to keep them as is.

PR #156, opened 2023-08-08
Add tests for ordering support

PaulLeCam:


packages/cli/test/composites.test.ts
```diff
@@ -62,7 +62,7 @@ describe('composites', () => {
       ])
       const output = create.stdout.toString()
       expect(output.includes('"version":"1.0"')).toBe(true)
-      expect(output.includes('"indices":{"Post"')).toBe(true)
+      expect(output.includes('"indices":{"')).toBe(true)
```
dbcfd, line 65: Post is present to verify that we are not generating empty indices.

dbcfd (changes requested):

packages/cli/test/composites.test.ts
PaulLeCam: The check still works with the presence of `"`, the test was failing because `Profile` was the first model returned, the test shouldn't assume any specific ordering here

dbcfd (approved):

PR #168, opened 2023-09-11
Add `nodes` entry point to GraphQL queries

PaulLeCam:
Support for this entry point is provided by the GraphQL library so adding support for it is trivial, this PR simply enables it.

### DIFF
--- a/packages/cli/test/__snapshots__/graphql.test.ts.snap
+++ b/packages/cli/test/__snapshots__/graphql.test.ts.snap
@@ -2,6 +2,12 @@
 
 exports[`graphql graphql:schema printing graphql schema succeeds 1`] = `
 "type Query {
+  """Fetches objects given their IDs"""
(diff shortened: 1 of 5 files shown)


PaulLeCam:
@stbrody it's mainly to avoid having to use multiple `node` entries in the query, see https://forum.ceramic.network/t/createindex-for-documentaccount-viewer/1279/5?u=paul for example

stbrody:
So I guess `node` entries are how we look up docs by StreamID?  So the user-visible feature here is a single query to load multiple documents by StreamID, is that right @PaulLeCam?

PaulLeCam:
Yes the `node` entry allows to load a specific doc by ID. It can be used multiple times in a single query so it's already possible to load multiple docs by ID at once, but the `nodes` entry allows to do so without having to provide a unique alias per doc to load.

PR #170, opened 2023-09-12
Add support for extra count queries with filters

PaulLeCam:
Adds entry-points for count queries in the `CeramicAccount` and `Query` objects.
~Tests should fail until Ceramic is updated with count queries support.~


packages/runtime/test/runtime.test.ts
```diff
@@ -219,6 +219,38 @@ describe('runtime', () => {
...
+      query {
+        ratingCount(filters: { where: { value : { lessThan: 3 } } })
+        viewer {
+          ratingListCount(filters: { where: { value : { greaterThan: 4 } } })
+        }
+      }
+      `,
+    )
+    expect(filteredCount.data?.ratingCount).toBe(1)
+    expect(filteredCount.data?.viewer?.ratingListCount).toBe(2)
```
stephhuynh18, line 253: I think this should be 1

stephhuynh18:
`'create and query post with comments` in both the `runtime` and `client` tests are failing because the snapshot is different. I actually think the snapshots are wrong and the tests are running correctly now. In the test we create two posts. We then create 3 comments for the second post. So the `commentCount` for the first post should be 0, and the `commentCount` for the second post should be 3. Please correct me if I am wrong.

packages/runtime/test/runtime.test.ts
PaulLeCam: Yes good catch, thanks! I updated it.

PaulLeCam:
@stephhuynh18 yes that's correct, I didn't notice at the time the counts were wrong, but now with the updated logic they needed to be updated indeed, this should be fixed now.
Once we have a stable release of the new Ceramic package I can update this PR with them and push a new release.

stephhuynh18 (approved):

PR #173, opened 2023-09-15
Adding Google Tag Manager

KGmajor:
Adding docusaurus plugin and google tag manager tag to trial Intercom and for better action tracking across the docs. 

✅ Built locally, no seen issues

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -18,6 +18,7 @@
   "dependencies": {
     "@docusaurus/core": "^2.4.1",
     "@docusaurus/plugin-content-docs": "^2.4.1",
+    "@docusaurus/plugin-google-tag-manager": "^2.4.1",
     "@docusaurus/preset-classic": "^2.4.1",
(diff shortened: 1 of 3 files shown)


website/package.json
```diff
@@ -18,6 +18,7 @@
   "dependencies": {
     "@docusaurus/core": "^2.4.1",
     "@docusaurus/plugin-content-docs": "^2.4.1",
+    "@docusaurus/plugin-google-tag-manager": "^2.4.1",
```
PaulLeCam, line 21: Seems unnecessary to add this plugin considering it's included in the [classic preset](https://docusaurus.io/docs/using-plugins#docusauruspreset-classic), no?

PaulLeCam (approved):

website/package.json
KGmajor: Just following directions from the docusaurus page 🤷🏽‍♀️. Unfortunately, no way to test the traffic locally.

PR #177, opened 2023-09-16
Mzk/sandbox 3

mzkrasner:
# [ComposeDB Sandbox - Adding DID filter] - #[Issue]

## Description

Adding in ability to filter by known DID on both profile and Posts models, as well as two new example queries

## How Has This Been Tested?

Builds locally


website/src/components/sandbox/profiles.js
```diff
@@ -14,6 +17,22 @@ const profileIndex =
   }
 }`
 
+const knownDid = 
+`query KnownDid {
+  node(id: "did:pkh:eip155:11155111:0x8071f6f971b438f7c0ea72c950430ee7655fabce") {
```
oed, line 22: What network is `11155111`?

mzkrasner: This is Sepolia testnet (I realized after generating the dataset this endpoint uses that I was logged into Sepolia instead of mainnet). If you'd rather this example show mainnet, I can go back and generate dummy data for a new address on mainnet?

PaulLeCam (approved):

PR #182, opened 2023-09-28
Add inverse relations to CeramicAccount

PaulLeCam:
Automatically adds views to the `CeramicAccount` object for inverse relations based on fields identified by the `@accountReference` directive, which should help cover some use-cases such as https://forum.ceramic.network/t/createindex-for-accountreference-did-fields/1288/1

Also adds support for setting the sync timeout on deterministic document creation for faster creation.


PaulLeCam:
@JustinaPetr I updated the docs relating to account references if you can have a look please?

packages/runtime/src/schema.ts
```diff
@@ -102,6 +106,20 @@ const connectionArgsWithAccount = {
...
+  fields: {
+    syncTimeout: {
+      type: GraphQLInt,
+      description: 'Maximum amount of time to lookup the stream over the network, in seconds',
+    },
+  },
+})
+
+type SetOptions = {
+  syncTimeout?: number
```
stbrody, line 120: Can we do this in a way that we don't need to update this for every new possible ceramic-side `CreateOpts` we want to expose?  Can we just expose a general purpose `createOpts` arg that just passes through to whatever `CreateOpts` the ceramic-level API exposes?

packages/runtime/src/schema.ts
```diff
@@ -102,6 +106,20 @@ const connectionArgsWithAccount = {
   },
 }
 
+const SetOptionsInput = new GraphQLInputObjectType({
+  name: 'SetOptionsInput',
+  fields: {
+    syncTimeout: {
+      type: GraphQLInt,
+      description: 'Maximum amount of time to lookup the stream over the network, in seconds',
```
stbrody, line 114: it sucks to have this defined in two places.  I'd rather just refer users to the js-ceramic API docs to see the docs on `CreateOpts.syncTimeoutSeconds`

packages/runtime/src/schema.ts
```diff
@@ -302,10 +328,23 @@ class SchemaBuilder {
                 { filters }: ConnectionFiltersArgument,
                 ctx,
               ): Promise<number> => {
-                if (filters != null) {
-                  assertValidQueryFilters(filters)
+                const query: BaseQuery = { model: model.id }
+                if (reference.type === 'account') {
```
stbrody, line 332: this looks like duplicated code with the above, can we add a helper for this?

packages/runtime/src/schema.ts
```diff
@@ -818,26 +857,49 @@ class SchemaBuilder {
...
-          throw new Error('Ceramic instance is not authenticated')
-        }
-        const document =
-          model.accountRelation.type === 'single'
-            ? await ctx.createSingle(model.id, input.content)
-            : await ctx.createDoc(model.id, input.content)
-        return { document }
-      },
-    })
+    if (model.accountRelation.type === 'single') {
```
stbrody, line 860: do we need to special-case single around this whole block like this?  If we can pass through the full `CreateOpts` object, then we can pass that down regardless of if its for a SINGLE or other type of document.  I actually think it would be better to expose the `CreateOpts` for all types of creates.  Ideally I'd love to expose the `UpdateOpts` on mutations as well while we're at it tbh

website/docs/api/runtime/schema.mdx
```diff
@@ -203,25 +209,33 @@ type CeramicAccount implements Node {
...
   isViewer: Boolean!
-  # The Post relation is a connection because the Post account relation is LIST
-  postList: PostConnection
-  # The Post relation is a single object because the Profile account relation is SINGLE
+  # The Meeting relation is a connection because the Meeting account relation is LIST
+  # The meetingList connection allows to access all the Meeting documents controlled by the account
+  meetingList: MeetingConnection
+  meetingListCount: Int!
+  # The otherOfMeetingList connection allows to access all the Meeting documents where the account DID is set as the value of the "other" field
+  otherOfMeetingList: MeetingConnection
```
stbrody, line 217: are these field names auto-generated?  So it's always `<field name of relation field>Of<model name of model with relation>Connection`?

That feels quite prescriptive to me, but I also can't see a way around if we want these accessors to be autogenerated...

In the future, if users don't like these autogenerated names, could we add an `@` directive to the composite to specify what the name of the autogenerated accessors should be?

stbrody (changes requested):
not a big deal but in general I think it's good practice to keep unrelated features in separate PRs.  For instance on this one I have more concerns about how we're exposing the CreateOpts than I do about the inverse relations, so if these were separate PRs the inverse relations one could already be merged.

packages/runtime/src/schema.ts
PaulLeCam: I think it's better to make these objects ComposeDB-specific and opt-in to use Ceramic types as needed, for example the update options in ComposeDB are unrelated to the Ceramic ones.

packages/runtime/src/schema.ts
PaulLeCam: Sure, do you have a link to the docs I could use here please?

packages/runtime/src/schema.ts
PaulLeCam: I'd rather avoid creating expectations that these inputs are a direct match for the Ceramic options. ComposeDB is an abstraction over Ceramic that has different needs, notably for updates there is an existing options object.

website/docs/api/runtime/schema.mdx
PaulLeCam: Yes it's possible to alias fields when performing queries, it's supported directly by GraphQL

packages/runtime/src/schema.ts
stbrody: Yeah I disagree.  I think it's fine for ComposeDB to have its own specific options for creates and updates, but I think porting over every ceramic-level option one by one is going to be a maintenance nightmare, lead to a lot of extra work for us, and lead to unexpected feature gaps where things work in Ceramic but are missed in ComposeDB.  I think it would be fine if the ComposeDB-level Update struct had a single field like `CeramicUpdateOpts` that passed through the UpdateOpts from the Ceramic API, I don't think we need to flatten the options up into the top level.  In fact I think that separation could be nice to make it very clear where the options are that are coming from the Ceramic API vs what is ComposeDB-specific.

PaulLeCam: With what you're suggesting we'd need to create the `CreateOpts` and `UpdateOpts` as GraphQL objects anyways, so to me this would be a bigger maintenance issue because that means any change on the Ceramic side would require an update on the ComposeDB side as well, we can't just "pass-through" options.

stbrody: gah, graphql always making things tricky!  What if we import the `CreateOpts` and `UpdateOpts` type definitions?  Can we then autogenerate the GraphQL objects off of those instead of having to manually re-define everything within them?  I just want to have a single place we can update and not have to try to manually keep the options in sync between js-ceramic and js-composedb

PaulLeCam: Type informations are not available at runtime so I don't think that's an option

packages/runtime/src/schema.ts
stbrody: Finally found them: https://developers.ceramic.network/reference/typescript/interfaces/_ceramicnetwork_common.CreateOpts.html

packages/runtime/src/schema.ts
stbrody: gah, I don't know enough about how the GraphQL type generation works, so I'll just have to take your word on it that there's no way to do this.  That's really sad to me, I really hate having these arguments duplicated in multiple places and needing to always update ComposeDB any time an option is changed in js-ceramic.  But if that's how it is then I guess that's how it is.

stbrody (approved):

packages/runtime/src/schema.ts
PaulLeCam: Thanks! I added the link

PR #191, opened 2023-11-07
Support interfaces creation

PaulLeCam:
Updates the devtools logic to support creating interface models, along with some changes to the runtime to make the tests pass.
Most of the added logic handles resolving dependencies between models and detecting cycles to ensure models can be loaded or created in order.


dbcfd:
This is a pretty large PR. Can we split it up? It looks like there's changes related to indexing, versioning, and the model definition, plus indices.

packages/devtools/test/composite.test.ts
```diff
@@ -555,13 +554,13 @@ describe('composite', () => {
...
+        const [postComposite, ratingComposite, noteComposite] = await Promise.all([
+          Composite.create({ ceramic, schema: postSchema }),
+          Composite.create({ ceramic, schema: ratingSchema }),
+          Composite.create({ ceramic, schema: noteSchema }),
+        ])
         const mergedComposite = postComposite.merge([ratingComposite, noteComposite]).toJSON()
-        expect(mergedComposite).toHaveProperty('indices.Post')
-        expect(mergedComposite).toHaveProperty('indices.Rating')
-        expect(mergedComposite).toHaveProperty('indices.Note')
+        expect(Object.keys(mergedComposite.indices ?? {})).toHaveLength(4)
```
dbcfd, line 563: This looks to originally have 3 indices, as checking exact property. Is 4 correct?

packages/devtools/test/composite.test.ts
```diff
@@ -739,21 +738,12 @@ describe('composite', () => {
...
-    expect(postID).toBeDefined()
-
-    const postAndCommentComposite = await Composite.create({
-      ceramic,
-      schema: createCommentSchemaWithPost(postID),
-    })
-    const commentID = postAndCommentComposite.modelIDs.find((id) => id !== postID)
-    expect(commentID).toBeDefined()
+    const composite = await Composite.create({ ceramic, schema: postSchema })
+    expect(composite.modelIDs).toHaveLength(2)
```
dbcfd, line 742: This needs a comment, since post schema actually has 2 models.

packages/runtime/src/schema.ts
```diff
@@ -377,15 +391,66 @@ class SchemaBuilder {
...
+    this._buildFiltersType(name, fields)
+    this._buildSortingType(name, fields)
   }
 
-  _buildDocumentObjectType({ model, definitions, name, fields }: BuildDocumentObjectParams) {
+  _buildDocumentObjectType({ model, definitions, name, fields }: BuildModelObjectParams) {
     this.#types[name] = new GraphQLObjectType<ModelInstanceDocument>({
       name,
-      interfaces: [definitions.nodeInterface],
+      interfaces: () => {
```
dbcfd, line 451: Does this need to be a function, seems like it was just an array before?

dbcfd (changes requested):
Is there a test where we pull by interface, not object?

packages/devtools/src/composite.ts
```diff
@@ -329,57 +258,29 @@ export class Composite {
...
-    definition.views = { models: modelsViews }
-    definition.indices = modelsIndices
-    const composite = new Composite({ commits, definition })
+    const { commits, views, ...definition } = await createIntermediaryCompositeDefinition(
+      params.ceramic,
+      abstractModels,
+    )
+    const composite = new Composite({
+      commits,
+      definition: {
```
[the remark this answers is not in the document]
PaulLeCam: They are part of the intermediary composite definition injected right below.

packages/devtools/test/composite.test.ts
PaulLeCam: Yes, the `postSchema` now also contains the `Comment` model creation that defines one index.

packages/devtools/test/composite.test.ts
PaulLeCam: What would the comment be please? That it's expected to have 2 models?

packages/runtime/src/schema.ts
PaulLeCam: Yes, that allows to defer the resolution until after all the types are defined

packages/devtools/test/composite.test.ts
dbcfd: Yep, that post schema includes comments. Or rename it to `postAndCommentSchema`

packages/devtools/test/composite.test.ts
dbcfd: Can we add back at least one check for the index matching, e.g. indices should contain Post.

PaulLeCam: OK, I replaced the length check by a snapshot

dbcfd (approved):

PR #192, opened 2023-11-07
Support interfaces queries

PaulLeCam:
Builds on https://github.com/ceramicstudio/js-composedb/pull/191

Updates the runtime logic to support interfaces and add tests for the happy paths.


packages/devtools/src/schema/resolution.ts
```diff
@@ -0,0 +1,340 @@
...
+      name: modelName,
+      model,
+      commitsPromise: loadCommits(ceramic, id),
+      indices: definition.indices ?? [],
+      views: definition.views,
+    }
+  }
+}
+
+function assertNoCircularDependency(
```
gvelez17, line 177: just one question, do we have to do this every time we load the model, or only when creating it?  Just don't want to have too much overhead if we have to test this every time

packages/devtools/src/schema/resolution.ts
```diff
@@ -0,0 +1,340 @@
...
+  for (const [name, resolve] of Object.entries(toResolve)) {
+    assertNoCircularDependency(toResolve, name)
+    if (resolve.requiredDependencies.length === 0) {
+      steps[0].push(resolve)
+    } else {
+      remainingModels.add(name)
+    }
+  }
+
+  // Add steps while there are remaining models to resolve
```
gvelez17, line 275: ok so this part requires that we already checked for circular dependencies, so it doesnt get stuck, right?  I wonder if there is a lighter weight check we could do rather than searching the whole tree for them, but anyway that's fine.

packages/runtime/src/schema.ts
```diff
@@ -709,22 +802,22 @@ class SchemaBuilder {
...
     const config: GraphQLInputFieldConfigMap = {}
     for (const [key, field] of Object.entries(fields)) {
       let type: GraphQLInputObjectType | undefined
-      if (field.type === 'reference' && (field as RuntimeReference).indexed) {
+      if (field.type === 'reference' && ((field as RuntimeReference).indexed || isInterface)) {
         if (field.refType === 'enum') {
           const enumType = this.#types[field.refName] as GraphQLEnumType
           type = createEnumValueFilterInput(enumType)
           this.#inputObjects[type.name] = type
         } else if (field.refType === 'node') {
```
gvelez17, line 817: most interfaces will be type node right?

packages/test-schemas/src/media.ts
```diff
@@ -1,36 +1,120 @@
...
+  collectionItems: [CollectionItem] @relationFrom(model: "CollectionItem", property: "itemID")
+  collectionItemsCount: Int! @relationCountFrom(model: "CollectionItem", property: "itemID")
+}
+
+type MyImage implements ImageMedia
+  @createModel(description: "A model for image objects")
+  @createIndex(fields: [{ path: ["src"] }])
+  @createIndex(fields: [{ path: ["width"] }])
+  @createIndex(fields: [{ path: ["height"] }]) {
+  src: String! @string(maxLength: 500)
```
gvelez17, line 68: can i have extra custom fields in here too, or no?

packages/test-schemas/src/media.ts
```diff
@@ -1,36 +1,120 @@
...
+  collectionItemsCount: Int! @relationCountFrom(model: "CollectionItem", property: "itemID")
+}
+
+type MyVideo implements VideoMedia
+  @createModel(description: "A model for video objects")
+  @createIndex(fields: [{ path: ["src"] }])
+  @createIndex(fields: [{ path: ["duration"] }])
+  @createIndex(fields: [{ path: ["width"] }])
+  @createIndex(fields: [{ path: ["height"] }]) {
+  src: String! @string(maxLength: 500)
```
gvelez17, line 92: do we have a test for interfaces with conflicting field definitions?  should that throw an error?

gvelez17:
Wow!  LGTM, excited to see this out!  Any performance implementation stuff can be a follow up anyway.  Is there a test that catches the error thrown for circular dependencies?  And just a couple questions in there, mainly clarity for us the great unwashed.  Thanks for getting this!

packages/devtools/src/schema/resolution.ts
PaulLeCam: It's only needed on creation because we don't support creating models with circular relations anyways, so here the logic is about making sure the composite the user attempts to define can actually be created.

packages/devtools/src/schema/resolution.ts
PaulLeCam: Yes I haven't tried to optimize the lookups but as it's only needed when creating composites and the number of models is likely pretty low, I don't think it's going to be an issue.

packages/runtime/src/schema.ts
PaulLeCam: Yes, `node` for a reference to a single document and `connection` for many

packages/test-schemas/src/media.ts
PaulLeCam: Yes! The models implementing interfaces can add extra fields as needed

packages/test-schemas/src/media.ts
PaulLeCam: No test yet as I've only focused on the "happy paths" so far but I'll add coverage for these kinds of checks

dbcfd (approved):

PR #196, opened 2023-11-15
feat: expose flag to determine whether a composite is indexed

dbcfd:


packages/cli/src/commands/composite/create.ts
```diff
@@ -23,12 +23,21 @@ export default class CreateComposite extends Command<Flags, { schemaFilePath: st
       char: 'o',
       description: 'path to the file where the composite representation should be saved',
     }),
+    index: Flags.boolean({
+      char: 'i',
+      description: 'If indexing should occur when creating composites (default=true)',
```
stbrody, line 28: Should the flag maybe be called "deploy" instead of "index"?  We should try to be consistent with whatever other naming conventions we are using elsewhere

stbrody, line 28: Some command line flag parsers are bad at letting boolean flags be set to false.  So let's definitely double-check if we haven't already that the one we're using supports setting `--flagName=false`

stbrody (commented):
basically looks good, adding Paul for naming convention questions

PaulLeCam (commented):
Looks OK to me but it misses docs and tests.

packages/cli/src/commands/composite/create.ts
dbcfd: So the flags come from oclif, and there are no args to the flag. It will be either `--flagName` or `--no-flagName`

```// flag with no value (-f, --force)
  force: Flags.boolean({
    char: 'f',                    // short character for flag
    default: true,                // default value if flag not passed (can be a function that returns a boolean)
    env: 'MY_NAME',               // default value to the value of an environment variable
    // boolean flags may be reversed with `--no-` (in this case: `--no-force`).
    // The flag will be set to false if reversed. This functionality
    // is disabled by default, to enable it:
    // allowNo: true
  }),```

File: packages/cli/src/commands/composite/create.ts [hunk 2]
```diff
@@ -23,12 +23,22 @@ export default class CreateComposite extends Command<Flags, { schemaFilePath: st
...
   }
 
   async run(): Promise<void> {
     try {
       this.spinner.start('Creating the composite...')
-      const composite = await createComposite(this.ceramic, this.args.schemaFilePath)
+      const composite = await createComposite(
+        this.ceramic,
+        this.args.schemaFilePath,
+        this.flags.deploy as boolean,
```
PaulLeCam, line 40: This shouldn't be needed, please make sure the flag is correctly typed

packages/cli/src/commands/composite/deploy.ts
```diff
@@ -17,13 +17,26 @@ export default class CompositeDeploy extends Command<
...
     try {
       this.spinner.start('Deploying the composite...')
       let composite: Composite | undefined = undefined
       if (this.stdin !== undefined) {
         const definition = JSON.parse(this.stdin) as EncodedCompositeDefinition
-        composite = await Composite.fromJSON({ ceramic: this.ceramic, definition, index: true })
+        composite = await Composite.fromJSON({
+          ceramic: this.ceramic,
+          definition,
+          index: this.flags.index as boolean,
```
PaulLeCam, line 38: This shouldn't be needed, please make sure the flag is correctly typed

PaulLeCam (changes requested):

packages/cli/src/commands/composite/deploy.ts
```diff
@@ -17,13 +17,26 @@ export default class CompositeDeploy extends Command<
     }),
   }
 
+  static flags = {
+    ...Command.flags,
+    index: Flags.boolean({
+      char: 'i',
+      description: 'Adds the composite to indexing on the ceramic node',
+      default: true,
```
stbrody, line 25: don't you need to set `allowNo` to enable the `--no-index` behavior?

website/docs/api/commands/cli.composite.md
```diff
@@ -126,6 +127,9 @@ ARGUMENTS
   
 OPTIONS
   -c, --ceramic-url        Ceramic API URL
+  -k, --did-private-key    DID Private Key (you can generate a fresh private key using composedb did:generate-private-key)
+  -i, --index              Adds the composite to indexing on the ceramic node (default)
+  --no-index               Prevents the composite from being indexed on the ceramic node. Useful if composite has already been deployed and indexed
```
stbrody, line 132: wait, what does it even mean to deploy a composite without indexing the models in it?

stbrody (commented):
my main question is around the behavior of the `--no-index` to the composite deploy command.  My understanding is that the very definition of "deploying" a composite means telling the node to index the constituent models.

website/docs/api/commands/cli.composite.md
stbrody: > Useful if composite has already been deployed and indexed

If the composite was already deployed and indexed, why would you be running `composite:deploy` again at all?

packages/cli/src/commands/composite/deploy.ts
dbcfd: It doesn't seem like it from the tests. Oclif just seems to support it automatically.

website/docs/api/commands/cli.composite.md
dbcfd: Possibly. They could have a CD process that just attempts to apply updates.

stbrody: what does that mean?  What do you mean by "apply updates"? Models can't be updated, so it's either deployed and indexed or not deployed and not indexed.  I still don't understand what it means to deploy without indexing.

PaulLeCam (changes requested):
As pointed out by Spencer, I don't get what is the purpose for deploying without indexing?
That would mean documents can be created but not queried on the node, which seems unlikely to be the wanted user experience?

dbcfd:
> As pointed out by Spencer, I don't get what is the purpose for deploying without indexing? That would mean documents can be created but not queried on the node, which seems unlikely to be the wanted user experience?

The desired user behavior is to update their composite with additional indices for querying. That requires create without indexing. However, they may also run deploy, since this is the steps specified in docs, so if they do that, need to do so without indexing.

website/docs/api/commands/cli.composite.md
stbrody: I just am still struggling to understand what deploying without indexing actually gains you.  It seems to me to be mostly a no-op.  If it doesn't do anything, then we shouldn't have people doing it.  If our docs need to be updated that's fine.

If there's some reason this is actually a useful feature other than "because our docs tell you to do it", then please let me know.  Maybe there's something I'm missing that this actually does which is useful.  But as of right now it doesn't seem to me like something there would ever be a legitimate reason to use.

dbcfd: I think that's a question for @PaulLeCam 

I'm not understanding why we have a deploy command.

PaulLeCam: > I'm not understanding why we have a deploy command.

I think this is the fundamental problem here, there seems to be a misunderstanding of how composites work.

There are 2 main parts to ComposeDB: the development tools that help manage the composites lifecycle and the runtime that generates a GraphQL schema using a compiled composite definition to interact with a Ceramic node.

In order for the runtime to work, it needs:
- A compiled composite definition, generated by `composite:compile` when using the CLI for example
- A Ceramic node indexing the models described in the composite, configured by running `composite:deploy` when using the CLI

So deploying a composite, whether by using the CLI or devtools libs, is needed to configure any Ceramic node to be usable by the ComposeDB runtime with an app's expected data model.
Creating a composite also deploys it to the given Ceramic node with the assumption that when developers create composites, they want to interact with the associated data as well. We can add an option not to index at this point as this PR does with the expectation developers will deploy the composite later, but the default should still be that a created composite is directly usable IMO.

Deploying a composite without indexing doesn't make sense to me, as it means the Ceramic node won't be able to handle queries from the ComposeDB runtime for the given composite. Do we have a specific use-case for this?

dbcfd: It's not a misunderstanding, it's a question about something you specifically reference

> So deploying a composite, whether by using the CLI or devtools libs, is needed to configure any Ceramic node to be usable by the ComposeDB runtime with an app's expected data model.
Creating a composite also deploys it to the given Ceramic node with the assumption that when developers create composites

As it is, create does the same functionality as deploy currently. This PR makes it so create doesn't index unless specified, so that deploy is not a no-op. However, if we deploy the composite once, we can't deploy it again without passing the `--no-index` flag. 

We could remove the deploy command, make the default behavior on create to index, and have a flag not to index for repeated updates to a composite. The question is, why should we retain the deploy command? It doesn't seem to have any functionality other than starting indexing, which create does anyways.

PaulLeCam: I just answered these questions in my previous comment, I don't really see how to explain it differently?

In a way deploying a composite is a bit like running `npm install` or similar to make sure an environment has the dependencies it requires to function. It is a no-op if the environment already has all the dependencies, but otherwise it's a necessary requirement to have the setup work as expected.

stbrody: > We could remove the deploy command, make the default behavior on create to index, and have a flag not to index for repeated updates to a composite. The question is, why should we retain the deploy command? It doesn't seem to have any functionality other than starting indexing, which create does anyways.

Yeah, it does seem like *either* we should just have the create command also control deploying/indexing the models with flags, *or* just always have a separate deploy command.

Personally I like keeping the separation between create and deploy as I do think those are conceptually different things and it seems clearer to me to keep them separate.  I also think it might be necessary for some workflows, such as when using a mix of pre-existing and net new models.  In that case you would want to use `composedb composite:from-model` to create a set of composites from the existing Models on the network you want to use, as well as use `composedb composite:create` to create a composite with all your new Models.  Then you probably want to use `composedb composite:merge` to merge them together, and then potentially update the merged composite to add any custom indices or views you want, and then only at the very end do you run `composedb composite:deploy` to actually deploy your final unified composite to the Ceramic node.

So my vote would be to make `composite:create` *not* deploy by default, so an explicit deploy is necessary.

I still don't think there's any reason to add a flag to deploy to allow deploying without indexing, because in my mind deploying == indexing.  The whole point of deploying a composite is to tell your Ceramic node to index the contained models.

packages/cli/src/commands/composite/create.ts
```diff
@@ -23,12 +24,22 @@ export default class CreateComposite extends Command<Flags, { schemaFilePath: st
       char: 'o',
       description: 'path to the file where the composite representation should be saved',
     }),
+    deploy: Flags.boolean({
+      char: 'd',
+      description:
+        'Deploy the composite to the ceramic node, which will start indexing on the composite',
+      default: true,
```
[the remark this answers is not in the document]
stbrody: Also, what does `composite:from-model` do?  That one seems even more like it definitely shouldn't auto-deploy since you're likely to want to add indices to it before deploying.  And if that one doesn't auto-deploy, then I feel like this shouldn't either just for consistency sake if nothing else

packages/cli/test/composites.test.ts
```diff
@@ -56,17 +62,37 @@ describe('composites', () => {
...
+        'test/mocks/composite.profile.post.schema',
+        `--did-private-key=${seed}`,
+      ])
+      const output = create.stdout.toString()
+      const def = JSON.parse(output) as EncodedCompositeDefinition
+      expect(def.version).toBe('1.1')
+      expect(Object.keys(def.version).length).not.toBe(0)
+      expect(def.aliases).toBeNull()
+      expect(def.views).toBeNull()
+      await expect(checkIfModelIndexed(ceramic, Object.keys(def.models)[0])).resolves.toBeFalsy()
```
stbrody, line 78: I'm surprised this test passes given that deploy is true by default  for `composite:create` in this version of the PR, and this test isn't explicitly passing `--no-deploy`.  What am I missing?

packages/devtools-node/src/fs.ts
```diff
@@ -25,9 +25,13 @@ export function getDirPath(path: PathInput): string {
...
  */
-export async function createComposite(ceramic: CeramicClient, path: PathInput): Promise<Composite> {
+export async function createComposite(
+  ceramic: CeramicClient,
+  path: PathInput,
+  deploy: boolean,
+): Promise<Composite> {
   const file = await readFile(getFilePath(path))
-  return await Composite.create({ ceramic, schema: file.toString() })
+  return await Composite.create({ ceramic, schema: file.toString(), index: !deploy })
```
stbrody, line 34: why is `index` the opposite of `deploy`?  Aren't those basically synonymous?

website/docs/api/commands/cli.composite.md
```diff
@@ -46,6 +46,10 @@ Create an encoded composite definition from GraphQL [Composite Schema](https://d
 
 You can find a detailed guide on the creation of Composites [here](https://developers.ceramic.network/docs/composedb/guides/data-modeling/composites)
 
+If updating your composite, run this command with `--no-deploy`. Your GraphQL
```
stbrody, line 49: I think we need to be clearer here about the types of updates we are talking about.  If you are updating your composite to add an entirely new model, then you do indeed want to deploy.  Also, attempting to index a model that's already indexed should be a no-op, so long as none of the custom indices have changed.  So I think we need to be more specific here that the issue is that you can't redeploy a composite to change the set of custom indices, to do that you need to manually connect to the SQL database and build the necessary indices yourself

website/docs/api/commands/cli.composite.md
```diff
@@ -57,6 +61,8 @@ OPTIONS
   -c, --ceramic-url        Ceramic API URL
   -k, --did-private-key    DID Private Key (you can generate a fresh private key using composedb did:generate-private-key)
   -o, --output             a path to file where the resulting encoded composite definition should be saved
+  -d, --deploy             Deploy the composite to the ceramic node, which will start indexing on the composite
```
stbrody, line 64:
```suggestion
  -d, --deploy             Deploy the composite to the ceramic node, which will cause the node to start indexing the Models contained within the composite
```

stbrody: suggested rewording for clarity

website/docs/api/commands/cli.composite.md
```diff
@@ -117,6 +123,12 @@ available on the Ceramic Node that yor DApp connects to. You can find a detailed
 guide on Composites' deployment
 [here](https://developers.ceramic.network/docs/composedb/guides/data-modeling/composites#deploying-composites)
 
+If updating your composite to add additional query fields, do not run this command.
```
stbrody, line 126: again, I think we need to be clearer with our language here about what exactly the problematic scenarios are and what the proper recourse is.

stbrody (changes requested):

packages/cli/src/commands/composite/create.ts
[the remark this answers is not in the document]
dbcfd: I was doing this just to maintain the existing behavior, so this change won't suddenly cause things to "stop working" for users that aren't calling deploy. 

I'm not entirely sure which way is better either.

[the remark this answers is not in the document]
stbrody: okay, I guess preserving existing behavior makes sense

packages/cli/src/commands/composite/from-model.ts
```diff
@@ -18,6 +19,12 @@ export default class CompositeFromModel extends Command<Flags> {
       char: 'o',
       description: 'path to the file where the composite representation should be saved',
     }),
+    deploy: Flags.boolean({
+      char: 'd',
+      description:
+        'Deploy the composite to the ceramic node, which will start indexing on the composite',
+      default: true,
```
stbrody, line 26: If I'm reading the code correctly, I believe this is a behavior change.  I'm pretty sure `composite:from-model` does not index currently.  I think this should probably default to false.

website/docs/api/commands/cli.composite.md
```diff
@@ -46,6 +46,10 @@ Create an encoded composite definition from GraphQL [Composite Schema](https://d
 
 You can find a detailed guide on the creation of Composites [here](https://developers.ceramic.network/docs/composedb/guides/data-modeling/composites)
 
+If updating your composite by specifying additional fields to filter on using the `createIndex` directive, run this 
+command with `--no-deploy`. Your GraphQL definition will still be updated, but Ceramic will not attempt to re-index your
```
stbrody, line 50: > will not attempt to re-index your composite

will not attempt to re-index *the Models in your composite*

website/docs/api/commands/cli.composite.md
stbrody: Did the language here get updated?  I like the language you used for the `create` command above (mod one small suggestion), can we re-use that?

packages/cli/src/commands/composite/from-model.ts
dbcfd: Good catch. That should default false.

qbig:
@PaulLeCam @stbrody The gist is we want to enable 2 workflows:
1. define models/composite from scratch & deploy and index them (pretty clear)
2. some other devs discover this model and would like to use it in their own application and as a result want to index these models and create new index definitions/composite (I think 2 is the confusing part)

stbrody (dismissed):
lgtm

packages/devtools-node/src/fs.ts
```diff
@@ -25,9 +25,13 @@ export function getDirPath(path: PathInput): string {
 /**
  * Create a Composite from a GraphQL schema path.
  */
-export async function createComposite(ceramic: CeramicClient, path: PathInput): Promise<Composite> {
+export async function createComposite(
+  ceramic: CeramicClient,
+  path: PathInput,
+  deploy: boolean,
```
PaulLeCam, line 31: That's a breaking API change, please set a default value or bump to v0.7 with docs updates.

packages/devtools-node/src/fs.ts
```diff
@@ -36,12 +40,12 @@ export async function createComposite(ceramic: CeramicClient, path: PathInput):
 export async function readEncodedComposite(
   ceramic: CeramicClient | string,
   path: PathInput,
-  index?: boolean,
+  deploy: boolean,
```
PaulLeCam, line 43: This is misleading, reading from JSON doesn't deploy the models to the node, please keep as-is.

packages/devtools/src/composite.ts
```diff
@@ -236,7 +236,7 @@ export type FromModelsParams = CompositeOptions & {
    * Whether to add the Models to the index or not. If `true`, the Ceramic instance must be
    * authenticated with an admin DID. Defaults to `false`.
    */
-  index?: boolean
+  index: boolean
```
PaulLeCam, line 239: Please keep this optional.

PaulLeCam (changes requested):

packages/devtools-node/src/fs.ts
dbcfd: It doesn't though. Composite create is the only command that creates (deploys) models to the node. `readEncodedComposite` deploys in the same manner that the deploy command deploys. For consistency between commands and SDK, everything that deploys (indexes) now has a deploy parameter.

PR #199, opened 2023-11-22
Remove container, add updated tag id

KGmajor:
Removing the container information that is used for installing heap and intercom. We'll move this over to the new docs.


KGmajor:
@JustinaPetr  Ready for review!

PR #203, opened 2023-12-11
Add clinet-side redirects

JustinaPetr:
## Description

Since the docs migration we haven't added the 1:1 redirects for the old composedb pages to the new site. This PR adds client-side redirects to the new site for relevant pages.


JustinaPetr:
@PaulLeCam I added the redirects for all of the old pages/versions. It can only tested in production though, but it should prevent 404s from happening from google search + any other links we have in the wild for older versions

PaulLeCam (approved):

PR #204, opened 2024-01-19
Support optional relations

PaulLeCam:
Following support for optional relations in Ceramic, this PR allows setting optional fields to `null` in the GraphQL updates and [apply logic to remove these fields](https://github.com/ceramicstudio/js-composedb/compare/fix/WS2-2819-optional-relations?expand=1#diff-00c5732531072980cd24487b37ff992c6fea8fcfe6c6a261c5447eab328e6260R65) for the JSON patch.

Opening as draft as the Ceramic dependencies will need to change from RC to stable versions, but should be OK to review already.


packages/runtime/src/loader.ts
```diff
@@ -124,7 +143,12 @@ export class DocumentLoader extends DataLoader<DocID, ModelInstanceDocument> {
       controller,
       model: model instanceof StreamID ? model : StreamID.fromString(model),
     }
-    const stream = await ModelInstanceDocument.create<T>(this.#ceramic, content, metadata, options)
+    const stream = await ModelInstanceDocument.create<T>(
+      this.#ceramic,
+      removeNullValues(content) as T,
```
dbcfd, line 148: What is the reason behind removing null values? There is a difference between `some_field: null` and `some_field` not being present. Especially from a verification standpoint, it lets us know they explicitly set that field to null.

PaulLeCam: The JSON schemas used in the models generated by ComposeDB don't use `null` values, if we send them they'll fail validation, as JSON schema differentiates between `required` properties and `null` property `type`.

stbrody: > The JSON schemas used in the models generated by ComposeDB don't use `null` values, if we send them they'll fail validation, as JSON schema differentiates between `required` properties and `null` property `type`.

Then wouldn't that be user error?  If the user wants to stop using a relation, they could just omit the field from the JSON they set the content to right?

Whatever we do here, we should try to be mindful that with things like Firehose and OrbisDB, the ComposeDB GraphQL tooling is no longer the only way that users will be interacting with Models and MIDs, so we should try to build things in a way that works outside of just the context of the CDB GraphQL APIs.

PaulLeCam: > Then wouldn't that be user error? If the user wants to stop using a relation, they could just omit the field from the JSON they set the content to, instead of setting it to null explicitly, right?

ComposeDB users don't need to send the full JSON content every time they make an update, they only need to set the fields they want changed, so here setting a field to `null` indicates it should be removed.

> Whatever we do here, we should try to be mindful that with things like Firehose and OrbisDB, the ComposeDB GraphQL tooling is no longer the only way that users will be interacting with Models and MIDs, so we should try to build things in a way that works outside of just the context of the CDB GraphQL APIs.

This PR is only a change to the ComposeDB runtime logic, it doesn't change the MIDs logic.

stbrody: > This PR is only a change to the ComposeDB runtime logic, it doesn't change the MIDs logic.

Ah yes, this seems like the right place then IMO

JulissaDantes (approved):
LGTM! Do you think that following this PR I should include a test case in #205  where an immutable field its updated to a null value?

PaulLeCam:
> LGTM! Do you think that following this PR I should include a test case in #205 where an immutable field its updated to a null value?

Thanks!
I think if we implement the change that the update mutation object doesn't include immutable fields, it shouldn't even possible to set the fields to `null` using the ComposeDB APIs, would it?

JulissaDantes:
> > LGTM! Do you think that following this PR I should include a test case in #205 where an immutable field its updated to a null value?
> 
> Thanks! I think if we implement the change that the update mutation object doesn't include immutable fields, it shouldn't even possible to set the fields to `null` using the ComposeDB APIs, would it?

No, you are right, just checking in case I was missing something

PR #205, opened 2024-01-24
Feature/ws2 2949 locking add graphql support

JulissaDantes:
We have to wait for the Ceramic side to be ready, but we could start reviewing this side in the meantime. Here we add the new directive to recognize immutable fields and make sure to support parsing it.


PaulLeCam (commented):
Looks great for the devtools part, but could you also update the runtime schema so that immutable fields are not present in the update inputs please?
The method generating these inputs is here: https://github.com/ceramicstudio/js-composedb/blob/main/packages/runtime/src/schema.ts#L866
I think that will also require some changes to the runtime definition to indicate what fields are immutable, here are the shared type definitions: https://github.com/ceramicstudio/js-composedb/blob/main/packages/types/src/index.ts

Let me know if you need more details or you want to sync on these changes some time!

packages/devtools/src/schema/parser.ts
```diff
@@ -482,6 +488,7 @@ export class SchemaParser {
     directives: Array<DirectiveAnnotation>,
   ): DefinitionWithReferences<ItemDefinition> {
     const required = isNonNullType(type)
+    const immutable = directives.some((item) => item.name === 'immutable')
```
PaulLeCam, line 491: I think we probably need an additional check somewhere that the directive is only used on models (having the `@createModel` directive present), as we don't support it on nested objects.

JulissaDantes: Please check my "feat:check if nested object" commit, I take the object directives to check for the `createModel` value.

packages/devtools/test/schema.test.ts
```diff
@@ -786,6 +786,90 @@ describe('schema parsing and compilation', () => {
...
+      createAbstractCompositeDefinition(`
+      interface TestInterface @loadModel(id: "interface ID") {
+        id: ID! @immutable
+      }
+
+      type TestModel @createModel(description: "Test model") {
+        text: String! @string(maxLength: 10)
+      }
+      `)
+    }).toThrow(`Unsupported immutable directive for id on nested object TestInterface`)
```
PaulLeCam, line 870: The error message is misleading here, as `TestInterface` is not a nested object.
Also that makes me wonder: has the interface implementation validation logic been updated to make sure the models implementing an interface with immutable fields also set these fields as immutable please?

PaulLeCam (changes requested):
LGTM other than the inherited immutable fields from the interfaces that could result in duplicate keys

PR #206, opened 2024-02-01
SET account relation support

PaulLeCam:
There's a bunch of unrelated changes due to the breaking changes in Ceramic APIs, as well as fixing some new linting errors caught by ESLint, but the main relevant changes are:

- Adding support for using the `SET` value as the `accountRelation` in the composite definition schema, as well as the `accountRelationFields`
- Generating queries based on the `SET` account relation, both for a unique value (with a dedicated input) and all values (similar to the `LIST` account relation)
- Generating mutations to create and update documents using the `SET` account relation
- Adding new `set[Model name]` mutations in addition to the now deprecated `create[Model name]` for `SINGLE` account relations, for consistency with the `SET` account relation mutations


packages/devtools/src/composite.ts
```diff
@@ -390,6 +391,18 @@ export class Composite {
     return Object.keys(this.#definition.models)
   }
 
+  /**
+   * Get the StreamID of the given model `alias` if present in the Composite.
+   */
+  getModelID(alias: string): string | null {
+    for (const [modelID, modelAlias] of Object.entries(this.#definition.aliases)) {
```
ukstv, line 398:
```
const found = Object.entries().find(([_, modelAlias]) => modelAlias === alias)
return found ? found.modelID : null
```

packages/devtools/src/formats/runtime.ts
```diff
@@ -344,14 +344,21 @@ export function createRuntimeDefinition(
...
+          case 'set':
+            // relation to single document in the set based on input
+            runtime.accountData[key] = { type: 'set', name: modelName }
+            // relation to all documents in the set
+            runtime.accountData[key + 'List'] = { type: 'connection', name: modelName }
+            break
+          case 'single':
+            runtime.accountData[key] = { type: 'node', name: modelName }
+            break
+          default:
```
ukstv, line 360: We have `UnreachableCaseError` to have exhaustiveness check.

ukstv (approved):
Minor not-important nitpicks. Can be ignored.

packages/runtime/test/runtime.test.ts
```diff
@@ -726,4 +727,89 @@ describe('runtime', () => {
...
+    })
+
+    const definition = composite.toRuntime()
+    expect(printGraphQLSchema(definition)).toMatchSnapshot()
+    const runtime = new ComposeRuntime({ ceramic, definition: composite.toRuntime() })
+
+    const createdPosts = await runtime.executeQuery<Record<string, { document: { id: string } }>>(
+      `
+      mutation CreatePosts(
+        $post1Input: CreatePostInput!,
```
JulissaDantes, line 755: Is this different from the setProfile in tests case `set profile with custom sync timeout`? Because I'm under the impression we changed the naming convention from `create` to `set` because we now do an upsert, but now I see 'create' profile being used.

Or is it that the `set profile with custom sync timeout` test case has something to do with set relationship type? If thats the case can you explain how.

packages/devtools/src/composite.ts
```diff
@@ -111,9 +107,14 @@ async function loadModelsFromCommits<Models = Record<string, StreamCommits>>(
         MODEL_GENESIS_OPTS,
       )
 
-      modelCommitsValues.filter(isSignedCommitContainer).map(async (commit: StreamCommits) => {
-        await assertSupportedReadModelController(model, commit as unknown as SignedCommitContainer)
-      })
+      await Promise.all(
+        modelCommitsValues.filter(isSignedCommitContainer).map(async (commit: StreamCommits) => {
```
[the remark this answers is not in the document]
PaulLeCam: Do you have more details why it causes troubles for Node please? For V8 using `async`/`await` is more optimized than hand-written promises.

packages/devtools/src/formats/runtime.ts
PaulLeCam: Yes I might add this at some point but I'd rather avoid importing more dependencies/APIs from Ceramic if that can be avoided.

packages/runtime/test/runtime.test.ts
PaulLeCam: Until now we've been creating `create[Model name]` and `update[Model name]` mutations for all models, whatever account relation type they use.
Now with the SET account relation, instead of `create[Model name]` we have `set[Model name]` which should be more explicit about the behavior and consistent with `unset[Model name]` to reverse/unindex.

For consistency, I think `set[Model name]` should be used as well for the SINGLE account relation because technically that's how it works, it can be treated more as an upsert than an insert, so I added `set[Model name]` for SINGLE account relation models, but kept the `create[Model name]` to avoid making it an immediate breaking change while developers transition to using `set[Model name]` instead. In a future release, we should stop generating these `create[Model name]` mutation for models using the SINGLE account relation.

These changes are only about the SET and SINGLE account relations, for the LIST account relation we still have (and we will continue to have) `create[Model name]` because each document is created uniquely.

Here in the tests, `set profile with custom sync timeout` is testing a model using the SINGLE account relation, while the added test covers the SET account relation. The `Post` model in this test uses the LIST account relation, but the test is mainly about the `Favorite` model which uses the SET account relation. Does that make sense?

JulissaDantes (approved):
LGTM! The only thing I would say is that I wish you would have done a PR addressing the adaptations to js-ceramic changes first, and then this one.

PaulLeCam:
> LGTM! The only thing I would say is that I wish you would have done a PR addressing the adaptations to js-ceramic changes first, and then this one. 

Thanks!
Yeah if I had known there would be breaking changes in Ceramic I would have done it differently, but it was unexpected to have to deal with it just to get this PR working.

PR #207, opened 2024-02-06
Extract loader and update logic for deterministic streams

PaulLeCam:
Following the discussion in https://github.com/ceramicnetwork/js-ceramic/pull/3103 and Spencer's fix of the multiquery logic on Ceramic's side, this PR improves the loader logic to handle deterministic streams and updates some queries to use deterministic lookups rather than index queries.
This PR also extracts the loader to a dedicated package so it can be more easily maintained and used by the ecosystem as needed.

https://github.com/ceramicstudio/js-composedb/pull/208 should be reviewed and merged first as the changes are included in this branch as well.


packages/loader/src/loader.ts
```diff
@@ -165,9 +179,23 @@ export class DocumentLoader extends DataLoader<DocID, ModelInstanceDocument> {
...
+   * Load a deterministic stream and add it to the cache.
+   */
+  async _loadDeterministic<T extends Record<string, any> = Record<string, any>>(
+    meta: GenesisMetadata,
+    options: CreateOpts = {},
   ): Promise<ModelInstanceDocument<T>> {
-    return (await super.load(id)) as ModelInstanceDocument<T>
+    const opts = { ...DEFAULT_DETERMINISTIC_OPTIONS, ...options }
+    const key = await this._getDeterministicKey(meta)
+    const stream = await this.load<T>({ ...key, opts })
```
stbrody, line 196: I don't really feel like I understand how DataLoader works well enough to know if this is safe or not.  It feels fishy to me, and I know the latency test you shared with me the other day was showing this code path to still have high latency (ie not be respecting the `opts` properly).  Are you sure this is going to pass the `CreateOpts` properly?

packages/loader/src/types.ts
```diff
@@ -0,0 +1,44 @@
+import type { GenesisCommit, LoadOpts } from '@ceramicnetwork/common'
+import type { ModelInstanceDocument } from '@ceramicnetwork/stream-model-instance'
+import type { CommitID, StreamID } from '@ceramicnetwork/streamid'
+
+export type DocID = CommitID | StreamID | string
```
stbrody, line 5: are there risks from letting this key be a StreamID or a string?  Feels like we would want to pick one and always coerce inputs into that?  What if you try to load the same stream, once via StreamID and again with a string, will they actually be treated as the same LoadKey and wind up in the same slot in the cache?

packages/runtime/src/context.ts
```diff
@@ -77,26 +67,10 @@ export type Context = {
     content: Content,
     options?: CreateOpts,
   ) => Promise<ModelInstanceDocument<Content>>
-  /**
-   * Update an existing document.
-   */
-  updateDoc: <Content extends Record<string, any>>(
```
stbrody, line 83: why are some of these methods removed from this class, but not others?

stbrody (commented):
I feel like there's a lot of context here that prevents me from being super thorough in reviewing this, but at a high level it generally seems fine.  A few specific questions inline

packages/loader/src/loader.ts
```diff
@@ -81,44 +75,45 @@ export function removeNullValues(content: Record<string, unknown>): Record<strin
...
-  controller?: string,
-): ModelInstanceDocumentMetadataArgs {
-  return {
-    controller,
-    model: model instanceof StreamID ? model : StreamID.fromString(model),
-  }
-}
-
-const tempBatchLoadFn: BatchLoadFn<DocID, ModelInstanceDocument> = () => Promise.resolve([])
+const tempBatchLoadFn: BatchLoadFn<LoadKey, ModelInstanceDocument> = () => Promise.resolve([])
```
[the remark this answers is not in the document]
PaulLeCam: IIRC this was needed because TS wouldn't allow calling `super()` with a function using `this`, but looks like it's since been fixed in TS so it's no longer needed, I updated the code.

packages/loader/src/types.ts
PaulLeCam: Yes, this is handled by this conversion function used by the cache: https://github.com/ceramicstudio/js-composedb/pull/207/files#diff-99f92e68738a8038a38d74537507f6b2e8831d31ba89f5f637a57c0c7d1dbff8R55-R57

packages/runtime/src/context.ts
PaulLeCam: I used to treat the loader as a purely internal API so the context would be the primary interface to interact with Ceramic, but now that the loader is extracted to a public package I think it's OK to use the `context.loader` directly from the GraphQL schema handlers.
These functions on the context were just proxies to the loader calls without any additional logic, so I think it's better to remove them and use the `context.loader` equivalents directly.

packages/loader/src/deterministic.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * Create a deterministic genesis commit matching the logic unsed internally by ModelInstanceDocument.
+ */
+export function createDeterministicGenesis(meta: GenesisMetadata): GenesisCommit {
+  const model = meta.model instanceof StreamID ? meta.model : StreamID.fromString(meta.model)
+  const header: GenesisHeader = { controllers: [meta.controller], model: model.bytes, sep: 'model' }
+  if (Array.isArray(meta.unique)) {
+    header.unique = new TextEncoder().encode(meta.unique.join('|'))
```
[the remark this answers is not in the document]
PaulLeCam: Yes, it provides the same output, that's tested in https://github.com/ceramicstudio/js-composedb/pull/207/files#diff-9416770bd4e5c5106a2eed5c312a1ed7f4edeb74ea3736db6b76e8b8daa3b050

I agree though it would be nice to have this directly available from the MID package, essentially there are 2 things needed here:

1. Create the genesis commit based on the input (this function)
2. Get the stream ID for the commit: https://github.com/ceramicstudio/js-composedb/pull/207/files#diff-0af4f6bd66935ecbd445ba22995dd687aeeee5468afdd46d060f8543a4a9a195R31

Do you see both functions as being part of the MID package or only the first one please? Shall we expose the stream ID from genesis function somewhere else?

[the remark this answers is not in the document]
stbrody: The former could certainly be made a public method in ModelInstanceDocument.

The latter, to get the StreamID from the genesis commit contents, that's not really specific to ModelInstanceDocuments.  I'd be okay with exposing a utility function from a js-ceramic package to do this, just need to think of what packages makes the most sense for it.  Maybe the StreamID package actually, though that would require pulling in the dag-jose codec into the streamid package which might bloat the package size...

packages/runtime/src/context.ts
stbrody: but then why not push all these functions down from the context into the loader?  Why do some stay here and others not? What's different between the ones that move vs the ones that stay here?  Can we treat them all the same?

stbrody (dismissed):

packages/runtime/src/context.ts
PaulLeCam: The context has additional logic, such as checking if the Ceramic instance has an authenticated DID or not and using the authenticated DID as controller for example.
I'd rather keep the loader rather low-level with the need to explicitly set the controller so the loader doesn't need to handle authentication at all, and leaving it up to the context or other higher level consumers to handle the logic they need, what do you think?

PR #208, opened 2024-02-09
Add support for set relation views

PaulLeCam:
This adds support for defining the `relationSetFrom` view in the definition schema and generating the views in the runtime schema.
I'll add more integration tests to show example queries and mutations but the snapshots already validate the schema is generated as expected.


packages/devtools/src/schema/parser.ts
```diff
@@ -450,6 +451,25 @@ export class SchemaParser {
...
+          }
+          const property = directive.args?.property as string | void
+          if (property == null) {
+            throw new Error(
+              `Missing property argument for @relationSetFrom directive on field ${fieldName} of object ${objectName}`,
+            )
+          }
+          return {
+            type: 'view',
+            required: false,
```
JulissaDantes, line 468: I thought we were going to set these fields as required

packages/devtools/src/schema/type-definitions.ts
```diff
@@ -35,8 +35,9 @@ directive @documentReference(model: String!) on FIELD_DEFINITION
 # Relation views
 
 directive @relationDocument(property: String!) on FIELD_DEFINITION
-directive @relationFrom(model: String!, property: String!) on FIELD_DEFINITION
+directive @relationFrom(model: String, property: String!) on FIELD_DEFINITION
```
JulissaDantes, line 38: Why don't we need the model here? Wouldnt this throw 'Missing model for ${source} relation view'?

JulissaDantes (commented):
I think it looks great, just a couple of questions

packages/devtools/src/schema/type-definitions.ts
PaulLeCam: I updated the logic to use the model defined in as the field type instead, so the `model` argument is no longer needed, I only kept it to avoid throwing errors in schemas that keep using it.

packages/devtools/src/schema/parser.ts
PaulLeCam: The view is a sort of "inverse" relation, basically here we lookup if there's a document with a model using the SET account relation that points to the current document, so it shouldn't be a required field because the relation may exist or not independently of the current document, if that makes sense?

JulissaDantes (approved):
LGTM!

PR #209, opened 2024-02-09
Only generate input objects with mutable fields

PaulLeCam:
Extra check to avoid GraphQL error with empty input field and adding the SET account relation fields to the immutable ones


packages/runtime/src/schema.ts
```diff
@@ -1311,11 +1326,18 @@ class SchemaBuilder {
...
-        id: { type: new GraphQLNonNull(GraphQLID) },
-        content: { type: new GraphQLNonNull(this.#inputObjects[`Partial${name}`]) },
-        options: { type: UpdateOptionsInput },
-      }),
+      inputFields: () => {
+        const inputFields: GraphQLInputFieldConfigMap = {
+          id: { type: new GraphQLNonNull(GraphQLID) },
+          options: { type: UpdateOptionsInput },
+        }
+        // It's possible content can't be updated if all fields are immutable
```
JulissaDantes, line 1334: Shouldnt we throw an error if an update mutation on an object with only immutable fields is attempted?

PaulLeCam: With this change the mutation input wouldn't have any `content` object so no field could be changed, but as we'll add the `shouldIndex` property to the `options` it should still be possible to toggle indexing.

JulissaDantes (approved):
LGTM!

PR #211, opened 2024-02-14
Feature/ws2 2953 unset graphql

JulissaDantes:
## Description
The users will now be able to call the `mutation Hide` special mutation to mark a stream as shouldIndex = false by passing the stream id.


packages/runtime/src/schema.ts
```diff
@@ -178,6 +178,27 @@ const UpdateOptionsInput = new GraphQLInputObjectType({
   },
 })
 
+const CeramicSignerInputType = new GraphQLInputObjectType({
+  name: 'CeramicSignerInput',
+  fields: {
+    isAuthenticated: { type: GraphQLBoolean },
```
PaulLeCam, line 184: What is the purpose of this input please?
Checking if the Ceramic instance has an authenticated DID should be handled internally by the context, not provided as input, if that's the goal here?

packages/runtime/src/schema.ts
```diff
@@ -178,6 +178,27 @@ const UpdateOptionsInput = new GraphQLInputObjectType({
...
+  name: 'HideOptionsInput',
+  fields: {
+    asDID: {
+      type: GraphQLDID,
+      description: 'DID for the signer(Optional)',
+    },
+    signer: {
+      type: CeramicSignerInputType,
+      description: 'Ceramic signer (Optional)',
+    },
```
PaulLeCam, line 198: Same as previous comment, this seems to be information we get from the context, not an input

PaulLeCam (changes requested):

packages/runtime/src/schema.ts
```diff
@@ -178,23 +178,12 @@ const UpdateOptionsInput = new GraphQLInputObjectType({
...
   name: 'HideOptionsInput',
   fields: {
-    asDID: {
-      type: GraphQLDID,
-      description: 'DID for the signer(Optional)',
-    },
-    signer: {
-      type: CeramicSignerInputType,
-      description: 'Ceramic signer (Optional)',
+    shouldIndex: {
```
PaulLeCam, line 209: Could you use https://github.com/ceramicstudio/js-composedb/blob/8cdf4ad26012cc10cf12d757c801d8000d35558f/packages/runtime/src/schema.ts#L152 directly so we have a single reference for this field please?

packages/runtime/src/schema.ts
```diff
@@ -178,23 +178,12 @@ const UpdateOptionsInput = new GraphQLInputObjectType({
...
 })
 
-const CeramicSignerInputType = new GraphQLInputObjectType({
-  name: 'CeramicSignerInput',
-  fields: {
-    isAuthenticated: { type: GraphQLBoolean },
-  },
-})
-
 const HideOptionsInput = new GraphQLInputObjectType({
```
PaulLeCam, line 202: Could you rename the input as we discussed please?
```suggestion
const EnableIndexingOptionsInput = new GraphQLInputObjectType({
```

PaulLeCam (changes requested):

packages/runtime/src/schema.ts
JulissaDantes: I had forgotten this, thanks for the reminder. Ready!

packages/runtime/src/context.ts
```diff
@@ -71,6 +71,10 @@ export type Context = {
     content: Content,
     options?: UpsertOptions,
   ) => Promise<ModelInstanceDocument<Content> | null>
+  /**
+   * Enable indexing for an existing document.
+   */
+  enableDocIndexing: (model: string, opts?: UpdateOpts) => Promise<void>
```
PaulLeCam, line 77: I think it's confusing to name the stream ID `model` here as it's the stream that gets updated, nothing to do with the model

```suggestion
  enableDocIndexing: (id: string, opts?: UpdateOpts) => Promise<void>
```

packages/runtime/src/context.ts
```diff
@@ -136,6 +140,18 @@ export function createContext(params: ContextParams): Context {
...
+      model: string,
+      opts?: UpdateOpts,
+    ): Promise<void> => {
+      const controller = getViewerID()
+      if (controller == null) {
+        throw new Error('Document can only be hidden with an authenticated account')
+      }
+
+      const doc = await loader.load<Content>({ id: model })
+      await doc.shouldIndex(false, opts)
```
PaulLeCam, line 153: The `shouldIndex` argument should be provided here so indexing can be disabled but also enabled back
```suggestion
      id: string,
      shouldIndex: boolean,
      opts?: UpdateOpts,
    ): Promise<void> => {
      const controller = getViewerID()
      if (controller == null) {
        throw new Error('Document can only be hidden with an authenticated account')
      }

      const doc = await loader.load<Content>({ id })
      await doc.shouldIndex(shouldIndex, opts)
```

packages/runtime/src/schema.ts
```diff
@@ -1363,6 +1374,24 @@ class SchemaBuilder {
         return { document: await ctx.loader.update(input.id, input.content, input.options) }
       },
     })
+
+    this.#mutations[`enableIndexing${name}`] = mutationWithClientMutationId({
+      name: `EnableIndexing${name}`,
+      inputFields: () => ({
+        id: { type: new GraphQLNonNull(GraphQLID) },
+        options: { type: EnableIndexingOptionsInput },
```
PaulLeCam, line 1382:
```suggestion
        shouldIndex { type: new GraphQLNonNull(GraphQLBoolean) },
```

packages/runtime/src/schema.ts
```diff
@@ -1363,6 +1374,24 @@ class SchemaBuilder {
...
+      }),
+      outputFields: () => ({
+        ...queryFields,
+        document: { type: this.#types[name] },
+      }),
+      mutateAndGetPayload: async (input: { id: string; options?: UpdateOpts }, ctx: Context) => {
```
PaulLeCam, line 1388:
```suggestion
      mutateAndGetPayload: async (input: { id: string; shouldIndex: boolean }, ctx: Context) => {
```

packages/runtime/src/schema.ts
```diff
@@ -1363,6 +1374,24 @@ class SchemaBuilder {
...
+        if (ctx.ceramic.did == null || !ctx.ceramic.did.authenticated) {
+          throw new Error('Ceramic instance is not authenticated')
+        }
+        return { document: await ctx.enableDocIndexing(input.id, input.options) }
```
PaulLeCam, line 1392:
```suggestion
        return { document: await ctx.enableDocIndexing(input.id, input.shouldIndex) }
```

packages/runtime/src/schema.ts
```diff
@@ -199,6 +199,17 @@ const UpdateOptionsInput = new GraphQLInputObjectType({
...
+  name: 'EnableIndexingOptionsInput',
+  fields: {
+    streamID: {
+      type: GraphQLString,
+      description: 'Reference a specific and unique event stream',
+    },
+    shouldIndexField,
+  },
+})
+
```
PaulLeCam, line 212: I don't think this is needed, the top-level fields of the mutation input should provide all the arguments
```suggestion
```

PaulLeCam (changes requested):
Thanks for the changes, I just noted a few things with using the term "model" for the doc ID and things that could be simplified I think

PaulLeCam (approved):
Thanks for the changes!

PR #212, opened 2024-02-16
Update docs for v0.7

PaulLeCam:
Mostly docs fixes and generated APIs updates, the main content change is about the directives to add the `SET` account relation and the `accountRelationFields`.

A small functional change as well is defaulting the `shouldIndex` to `true` when calling the `set<Model>` mutations, so unless the `shouldIndex` option is explicitly set to `false`, the document should be indexed, I think this behavior would better match the expectations.

### DIFF
--- a/website/docs/api/modules/runtime.md
+++ b/website/docs/api/modules/runtime.md
@@ -33,17 +33,13 @@ ___
 | Name | Type | Description |
 | :------ | :------ | :------ |
 | `ceramic` | `CeramicAPI` | Ceramic client instance used internally. |
-| `createDoc` | \<Content\>(`model`: `string`, `content`: `Content`) => `Promise`\<`ModelInstanceDocument`\<`Content`\>\> | Create a new document with the given model and content. |
 | `getViewerID` | () => `string` \| ``null`` | ID of the current viewer (authenticated DID), if set. |
 | `isAuthenticated` | () => `boolean` | Returns whether the Ceramic client instance used internally is authenticated or not. When not authenticated, mutations will fail. |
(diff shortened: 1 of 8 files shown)


website/docs/api/modules/runtime.md
```diff
@@ -33,17 +33,13 @@ ___
...
-| `upsertSet` | \<Content\>(`model`: `string`, `unique`: `string`[], `content`: `Content`, `options?`: `CreateOpts`) => `Promise`\<`ModelInstanceDocument`\<`Content`\>\> | Create or update a document using the SET account relation with the given model, content and unique fields value. |
-| `upsertSingle` | \<Content\>(`model`: `string`, `content`: `Content`, `options?`: `CreateOpts`) => `Promise`\<`ModelInstanceDocument`\<`Content`\>\> | Create or update a document using the SINGLE account relation with the given model and content. |
+| `upsertSet` | \<Content\>(`model`: `string`, `unique`: `string`[], `content`: `Content`, `options?`: `UpsertOptions`) => `Promise`\<`ModelInstanceDocument`\<`Content`\> \| ``null``\> | Create or update a document using the SET account relation with the given model, content and unique fields value. |
+| `upsertSingle` | \<Content\>(`model`: `string`, `content`: `Content`, `options?`: `UpsertOptions`) => `Promise`\<`ModelInstanceDocument`\<`Content`\> \| ``null``\> | Create or update a document using the SINGLE account relation with the given model and content. |
```
JulissaDantes, line 42: Now that the the #211 is merged we should include `hideDoc`, no?

PaulLeCam: Yes it's generated automatically when building the docs, I checked manually so it's also added in the diff now

m0ar (commented):
Psst, missing information about the `@immutableFields` directive in the schema definition language. Seems it's only mentioned in the devtools :angel:

PR #213, opened 2024-02-16
Update dependencies

PaulLeCam:
Main change is having the add the `--strip-leading-paths` flag to preserve the current output behavior, as the SWC CLI changed the default output structure


JulissaDantes (approved):
LGTM!

PR #214, opened 2024-02-21
Fix test environment

PaulLeCam:
I think the issue with intermittent flaky tests is caused by the in-memory anchor service, adding these flags to the Ceramic config seems to solve the issue: https://github.com/ceramicstudio/js-composedb/compare/chore/social-network-tests?expand=1#diff-559e5a501a171aa2dd682776ad9efdc0a463a13cd8a4ee824e3811895df25a71R48-R49

Other changes include fixing some linting warnings and adding more tests for the runtime behavior, I'll keep adding more tests in another PR but it would be good to merge this one already I think to check if it indeed solves the issue we've seen.


packages/runtime/test/social-network.test.ts
```diff
@@ -0,0 +1,292 @@
...
+    await createPostsAndTags(user2, {
+      post1: { content: { date: '2024-02-10T01:10:00Z', text: 'Hello there' } },
+      post2: { content: { date: '2024-02-13T09:30:00Z', text: 'Hello again' } },
+      post3: { content: { date: '2024-02-16T06:15:00Z', text: 'Whats up?' } },
+      tag1: { content: { name: 'Hello' } },
+      tag2: { content: { name: 'Random' } },
+    })
+    await createPostsAndTags(user3, {
+      post1: { content: { date: '2024-02-02T01:10:00Z', text: 'Hello' } },
+      post2: { content: { date: '2024-02-05T09:30:00Z', text: 'Ceramic rocks' } },
```
JulissaDantes, line 138: Nice text values!

JulissaDantes (changes requested):
Everything looks good, but I would like to see a tests case for trying to maybe like a post twice or something that will not be the expected use for set relations, to make sure all the rules are enforced.

JulissaDantes (approved):
LGTM!

PR #215, opened 2024-02-26
Add option to ignore deterministic docs with empty contents

PaulLeCam:
This should be more in line with the expected behavior when looking up deterministic streams to check if a relation exists

### DIFF
--- a/packages/loader/src/loader.ts
+++ b/packages/loader/src/loader.ts
@@ -27,7 +27,6 @@ export type CreateOptions = CreateOpts & {
 }
 
 export type DeterministicLoadOptions = CreateOpts & {
-  controller?: string
   ignoreEmpty?: boolean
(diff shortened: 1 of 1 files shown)


packages/loader/src/loader.ts
```diff
@@ -26,7 +26,9 @@ export type CreateOptions = CreateOpts & {
   shouldIndex?: boolean
 }
 
-export type DeterministicLoadOptions = CreateOptions & {
+export type DeterministicLoadOptions = CreateOpts & {
+  controller?: string
```
JulissaDantes, line 30: This new `controllers` property doesnt get used, is this for a future development?

PaulLeCam: No it's not needed actually, good catch!

JulissaDantes (approved):
LGTM!

PR #217, opened 2024-03-12
feat: adds document:list command

nathanielc:
The command works much like the model:list command but has two differences:

1. It list documents within a specific model instead of the metamodel
2. It uses the local node instead of an indexer node.

This way users can have a simple way to read documents they have created via the CLI.

## How Has This Been Tested?

Created a model with various documents and then ran the list command in json and table modes.

### DIFF
--- a/packages/cli/src/commands/document/list.ts
+++ b/packages/cli/src/commands/document/list.ts
@@ -1,0 +1,139 @@
+import { Command, type CommandFlags } from '../../command.js'
+import { Args, Flags, ux } from '@oclif/core'
+import Table, { Cell } from 'cli-table3'
+import { Edge, Page, PageInfo, StreamState } from '@ceramicnetwork/common'
+import terminalSize from 'terminal-size'
+import { CeramicClient } from '@ceramicnetwork/http-client'
+import { StreamID } from '@ceramicnetwork/streamid'
+
+type PartialDocumentDefinition = {
+  id: string
+  controller: string,
+  content: any
+}
+
+type DocumentListFlags = CommandFlags & {
+  table?: boolean
+}
+
+export default class DocumentList extends Command<DocumentListFlags, { model: StreamID }> {
+  fetchedFields: Array<PartialDocumentDefinition> = []
+  lastLoadedPageInfo: PageInfo | null = null
+
+  static description = 'load a list of documents for a model with pagination'
+
+  static args = {
+    model: Args.custom({
+      required: true,
+      description: 'StreamID of the model containing documents',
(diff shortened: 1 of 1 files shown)


stbrody:
What's the behavior if the number of matching documents is large?  Is there any pagination?  I'd like to avoid another command like listing pinned streamids in the Ceramic CLI that just hangs for a long time then returns all the pinned streams in a single http response.

PaulLeCam:
Thanks for the PR but I'd rather avoid adding interactions commands here, IMO the GraphQL server should be the go-to solution for interacting with the composite.
Beyond pagination as mentioned by Spencer, we might have requests to support filtering, relations, mutations, etc. that basically end-up duplicating what is already provided by GraphQL.

nathanielc:
My aim is to provide an initial experience via the CLI to get started with ComposeDB. Currently its possible to create a model and then create instance of that model, however its not possible to read any documents back out of the server. It seems odd to not be able to do a read from the CLI and instead change your workflow to use the GraphQL server. I am not looking for a in depth query experience on the CLI rather a basic I can validate that the data I just wrote exists.

As for pagination, this command has pagination. It uses the same pagination as the `model:list` command, i.e. it waits for user input before requesting another page of data.

PaulLeCam:
OK I guess the confusion here comes from the fact there are the `document` commands that really only exist from legacy reasons, I think we should remove them and only use GraphQL for interactions.
Maybe instead of adding to the `document` commands, we could add a `graphql:execute` command to execute queries directly from the CLI, what do you think?

nathanielc:
I see a lot of value in a DX that is CLI based. Having to switch to a browser in order to complete basic tasks like reading data is not ideal. I'd prefer a strong CLI based workflow. However this is getting bigger in scope than my PR.

nathanielc:
Would a `graphql:execute` command be able to do mutations as well as queries? The more I think about it the more a single graphql command makes sense to replace all of the `document` commands. 

Would the `graphql:execute` command also use the DID from the environment as the current session? I might give this a try if its something you think would make sense.

PaulLeCam:
Yes I agree it would be nice to have better CLI functionalities, but I think that should be driven by actual customer demands, otherwise we're just adding more code to test and maintain without a real need.

Having a single `graphql:execute` instead of all the `document` commands would be easier to maintain IMO. I expect it should work with minimal logic and yes supporting mutations if a DID key is provided, assuming we implement the necessary logic.

nathanielc:
@PaulLeCam Sounds good I created https://github.com/ceramicstudio/js-composedb/pull/218 and its working well and was a very simple addition.

PaulLeCam:
Thanks!

PR #219, opened 2024-03-15
feat: add github workflow to publish composedb-cli image

3benbox:
Publish an image for composedb-cli so that we can use the did tools to convert private keys to dids in keramik.

### DIFF
--- a/.github/workflows/docker-cli.yaml
+++ b/.github/workflows/docker-cli.yaml
@@ -42,3 +42,9 @@ jobs:
           build-args: |
             VERSION=${{ steps.package-version.outputs.version }}
           tags: ${{ env.DOCKER_IMAGE }}:${{ steps.package-version.outputs.version }}
+
+      - name: Push Docker image with latest tag (if default branch)
+        if: github.event.inputs.branch == 'main'
+        run: |
+          docker tag ${{ env.DOCKER_IMAGE }}:${{ steps.extract_version.outputs.version }} ${{ env.DOCKER_IMAGE }}:latest
+          docker push ${{ env.DOCKER_IMAGE }}:latest
--- a/.github/workflows/docker-cli.yaml
+++ b/.github/workflows/docker-cli.yaml
@@ -1,0 +1,44 @@
+name: Build and Push composed-cli Docker Image
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Branch name"
+        default: "main"
+        required: true
+
+env:
+  DOCKER_IMAGE: ceramicnetwork/composedb-cli
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-cli.yaml
```diff
@@ -0,0 +1,28 @@
+name: Build Docker Image
+
+on:
+  workflow_dispatch:
+    inputs:
+      COMPOSEDB_CLI_VERSION:
+        description: 'ComposeDB CLI version'
+        default: 'latest'
```
smrz2001, line 8: Can we detect the CLI version and use that? I feel like the `latest` tag should apply to all `main` builds and the actual CLI version tag should apply when a new release is created.

.github/workflows/docker-cli.yaml
```diff
@@ -0,0 +1,28 @@
...
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - name: Build Docker image
+      run: |
+        docker build --build-arg COMPOSEDB_CLI_VERSION=${{ github.event.inputs.COMPOSEDB_CLI_VERSION }} -t ceramicnetwork/composedb-cli:${{ github.event.inputs.COMPOSEDB_CLI_VERSION }} .
+
+    - name: Push Docker image
+      run: |
+        docker push ceramicnetwork/composedb-cli:${{ github.event.inputs.COMPOSEDB_CLI_VERSION }}
```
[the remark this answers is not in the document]
3benbox: Added!

.github/workflows/docker-cli.yaml
3benbox: Updated to take a branch and version from that branch.

Unsure how the releases are created.

smrz2001 (approved):
Can figure out how to only tag the release after a release commit later, but otherwise LGTM.

.github/workflows/docker-cli.yaml
```diff
@@ -0,0 +1,44 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          build-args: |
+            VERSION=${{ steps.package-version.outputs.version }}
+          tags: ${{ env.DOCKER_IMAGE }}:${{ steps.package-version.outputs.version }}
```
smrz2001, line 44: Can you add the `latest` tag here as well since the image is tracking `main`?

3benbox: Added!

smrz2001 (approved):
🚀

PR #226, opened 2024-07-17
v0.8 release

PaulLeCam:
Mainly needed to be in sync with the v6.x of js-ceramic packages.
The only other change is removing the `document` commands from the CLI to interact with documents, as the `graphql:execute` command should be used instead.


oed (commented):
Lgtm. Do we know if anyone is using the `document` commands? If a lot of people rely on them it might not make sense to remove them?

PaulLeCam:
I don't know if anyone relies on them, they were mainly ported from the Glaze CLI and left there, but for all interactions in ComposeDB we recommended using GraphiQL.
One problem with these `document` commands is that they don't support deterministic streams, so their use-case is limited, while the added `graphql:execute` command supports all the interactions.

JulissaDantes:
Can we wait to include #227 in this release?

JulissaDantes (approved):
LGTM!

PaulLeCam:
Thanks!

PR #229, opened 2024-07-26
feat: update _parseListType

JulissaDantes:
Description
In the current version a field declared as uniqueListValue: [Int] @list(maxLength: 5) @immutable is not included in the ImmutableFields. This PR allows to identify these kinds of declarations.

This is the working CI version of #227

### DIFF
--- a/packages/devtools/src/schema/parser.ts
+++ b/packages/devtools/src/schema/parser.ts
@@ -326,7 +326,6 @@ export class SchemaParser {
     for (const [key, value] of Object.entries(objectFields)) {
       const directives = getDirectives(this.#schema, value)
 
-      const immutable = directives.some((item) => item.name === 'immutable')
       const [innerType, required] = isNonNullType(value.type)
         ? [value.type.ofType, true]
         : [value.type, false]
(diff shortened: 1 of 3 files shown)


packages/devtools/src/schema/parser.ts
```diff
@@ -326,6 +326,7 @@ export class SchemaParser {
     for (const [key, value] of Object.entries(objectFields)) {
       const directives = getDirectives(this.#schema, value)
 
+      const immutable = directives.some((item) => item.name === 'immutable')
```
PaulLeCam, line 329: Does the directive check need to be here, rather than part of the `_parseListType()` method?

PaulLeCam (approved):
LGTM, thanks for the changes!

packages/devtools/src/schema/parser.ts
JulissaDantes: could be in either place I think I put it here wanted it to look homogeneous to another parse method, but Ill put it directly in the method

JulissaDantes: Done!

PaulLeCam (approved):
thanks!

